PR #11, opened 2017-03-24
add graphics.Dimension to legacy.ENTITY_WRAPPERS

niki-sp:
The legacy factory returns DEFAULT_WRAPPER for DIMENSION entity. Proper wrapper class is graphics.Dimension.


mozman:
It is a conscious decision to not support DIMENSION yet.

YunbiaoCao:
Now, I'm working on one CAD project and it's a good starting with this lib except the dimension.  So, could you let me know why we don't support dimension? 

Thanks!

mozman:
Because WE have no time for that.

PR #29, opened 2018-02-16
Added missing ezdxf.audit line to setup.py

csachs:
I was just about to prepare a package for [conda-forge](https://conda-forge.org/) (would you be interested to be added as maintainer?) as I spotted that the `dxfaudit` is currently not working, as `ezdxf.audit` is not added to the package files (it is missing from the PyPI source distribution as well).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,7 @@ setup(
               'ezdxf.sections',
               'ezdxf.templates',
               'ezdxf.tools',
+              'ezdxf.audit',
               ],
     package_data={'ezdxf': ['templates/*.dxf',
                             'pp/dxfpp.html',


mozman:
Thank you, added line without merge

mozman:
Added a tox configuration, and added integration_tests and associated test data to the source package, so the tox command should run all tests and integration test. I hope this is helpful.

csachs:
Wanted to update the package for conda-forge, but spotted that the `dxfaudit` line is gone again (by commit 6be5bac888259629ec4938562104984b3b5da97e), while the code of the audit module is present. Was this intended?

mozman:
Yes, `dxfaudit` command was deliberately removed. The audit sub-package will be extended or removed in the future, but the current state it is not very useful.

csachs:
Thank you for the clarification.

Issue #37, opened 2018-07-23
Why drawing arcs clock-wisely when adding hatches?


DanielPompermayer:
Why is drawing arcs clock-wisely default when adding it to a hatch boundary EdgePath, while counter-clock-wisely is the default when adding an arc to some layout?

Since layout.add_arc hasn't the is_counter_clockwise parameter, we cannot write some functions to draw some complex and useful objects and use them to fill those objects.

mozman:
The DXF standard does not define a counter clockwise flag for the ARC entity.

http://help.autodesk.com/view/OARX/2018/ENU/?guid=GUID-0B14D8F1-0EBA-44BF-9108-57D8CE614BC8

DanielPompermayer:
Yes, you're right. But despite not having any flag, it actually draws an arc counter counter-clock-wisely.

The trouble, I think, is in "is_counter_clockwise" parameter of the "add_arc" method when drawing the edge_path in the boundary context manager. This parameter has the default parameter as 0, which makes the method drawing the arcs clock-wisely.

DanielPompermayer:
For ensuring compatibility, it would not be a good idea to change the default value of "is_counter_clockwise" in the edge_path boundary context manager. So I'm submitting a [pull request](https://github.com/mozman/ezdxf/pull/38) for adding this parameter in GraphicsFactory.add_arc method.

PR #38, opened 2018-07-26
Added is_counter_clockwise in GraphicsFactory.

DanielPompermayer:
Originally, GraphicsFactory.add_arc hasn't the is_counter_clockwise parameter, thus we cannot write some functions to draw some complex and useful objects and use those functions to fill those objects using hatch. This change, while ensuring the compatibility, will allow users to do that.

Discussion in [issue 37](https://github.com/mozman/ezdxf/issues/37).


mozman:
Thank you for your contribution, added your changes as patch to the develop branch.

PR #43, opened 2018-09-03
Prevent missing linetype error in table addon.

chr15m:
DXFs with tables generated without this fix cause the following error on load:
The following error was encountered while reading
in LINE starting at line 212132:
Bad linetype name None
Invalid or incomplete DXF input -- drawing discarded.

https://groups.google.com/forum/#!topic/python-ezdxf/iyJa0_ydgqc


mozman:
Applied patch to the development branch, thank you for your contribution.

chr15m:
Thank you for a very solid piece of software.

PR #50, opened 2018-10-05
Fix table addon header twisted polygon.

### LINKED ISSUE #49: table addon render_cell_background is creating twisted polygon
Looks like this:

table-fix-old

Should look like this:

table-fix-new

@mozman I have a fix for this. Will issue a PR against `develop` branch.

chr15m:
Fixes #49.


mozman:
thank you, again!

PR #56, opened 2018-11-03
"coding: utf-8" magic comments (PEP 263)

speleo3:
Some source files have UTF-8 characters. Python 2 fails with `SyntaxError: Non-ASCII character`. Adding magic encoding comments, like suggested by [PEP 263](https://www.python.org/dev/peps/pep-0263/), fixes the problem.


mozman:
Thank you for your contribution, but I don't accept pull requests to the master branch.

PR #61, opened 2018-12-30
Add encoding to files containing utf-8 characters.

chr15m:
@mozman this fixes a bunch of issues with file encodings. Basically I found every file where there were utf-8 characters and added an encoding declaration. I was running into SyntaxErrors when importing ezdxf on the Heroku platform and this PR fixess them.

Conflicts:
	ezdxf/legacy/graphics.py
	ezdxf/modern/dimension.py
	ezdxf/pp/dxfpp.py
	ezdxf/tools/complex_ltype.py
	tests/test_compressed_string.py


mozman:
Hi, thank you for you contribution, but I don't accept pull requests for the master branch. 

This problem is also only Python 2 related and Python 2 support is already removed in the actual develop branch, the only branch this fix makes sense is the 0.8.10 branch which is the bugfix branch for the still Python 2 compatible 0.8.x version.

chr15m:
@mozman no problem, should I make a new PR against `0.8.10`?

mozman:
Yes, please!

PR #72, opened 2019-02-25
Correct wrong DXF version in doc.

mizuno-gsinet:
According to AutoCAD [2009](https://damassets.autodesk.net/content/dam/autodesk/www/developer-network/platform-technologies/autocad-dxf-archive/acad_dxf_2009.pdf), [2010](http://images.autodesk.com/adsk/files/acad_dxf1.pdf) and [2011](http://images.autodesk.com/adsk/files/acad_dxf2.pdf) DXF references, MESH entity seems to be introduced in AutoCAD 2010 DXF.

Although it is not actually documented in the 2010 reference, according to [CONVTOMESH](https://knowledge.autodesk.com/support/autocad/learn-explore/caas/CloudHelp/cloudhelp/2019/ENU/AutoCAD-Core/files/GUID-810D73BD-1C40-4985-AF31-60D9857F6E64-htm.html) command reference quoted below, it should be available since 2010.

>Use this method to convert 3D faces (3DFACE) and legacy polygonal and polyface meshes (from Release 2009 and earlier). You can also convert 2D objects such as regions and closed polylines.

### DIFF
--- a/docs/source/dxfentities/mesh.rst
+++ b/docs/source/dxfentities/mesh.rst
@@ -3,7 +3,7 @@ Mesh
 
 .. class:: Mesh(GraphicEntity)
 
-    Introduced in DXF version R13 (AC1012), dxftype is MESH.
+    Introduced in DXF version R2010 (AC1024), dxftype is MESH.
 
     3D mesh entity similar to the :class:`Polyface` entity. Create :class:`Mesh` in layouts and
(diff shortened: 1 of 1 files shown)


mozman:
Hi, thank you for your contribution but:

1. What Autodesk says and what AutoCAD does are different things, fact is that TrueView and BricsCAD accept MESH entities starting at least DXF R2000.
2. I don't accept pull requests into the master branch

PR #107, opened 2019-12-21
bug fix of encoding in readfile()

KatsuhiroMorishita:
Since the character code was ignored even if specified, the program was modified.

### DIFF
--- a/ezdxf/filemanagement.py
+++ b/ezdxf/filemanagement.py
@@ -118,12 +118,16 @@ def readfile(filename: str, encoding: str = None, legacy_mode: bool = False, fil
         raise IOError("File '{}' is not a DXF file.".format(filename))
 
     info = dxf_file_info(filename)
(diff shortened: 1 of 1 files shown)


mozman:
If you set an unsupported codepage as it is possible in your code, ezdxf will save your DXF with `cp1252` encoding, all supported codepages I know can be founds here: https://github.com/mozman/ezdxf/blob/master/ezdxf/tools/codepage.py

If you know more supported codepages (by AutoCAD), let me know.

mozman:
Added a bugfix for your problem, without setting an unsupported encoding, but this case should not happen if the DXF file is saved by AutoCAD. If you use DXF R2007 the encoding would always be `utf-8`, regradless the `$DWGCODEPAGE` setting. 

If you have a DXF file with another encoding setting `ANSI_xxxx` as in encoding.py listed which can be opened by AutoCAD, let me know and please add the required Python encoding settings as well and maybe you can provide an example DXF file

Thank you,
Manfred

PR #112, opened 2020-01-13
Construction Line intersect bug

ScottHilton:
Sometimes ConstructionLine has_intersect returns false when it should be true.

My construction line is horizontal on the line y=248.725, the point returned by self.ray.intersect has y=248.72500000000002. The point is therefore not in the bounding box and has_intersect returns true.

This change fixes my bug and should cover all edge cases I can think of.

### DIFF
--- a/ezdxf/math/line.py
+++ b/ezdxf/math/line.py
@@ -116,11 +116,23 @@ class ConstructionRay:
         ray2 = other
         if not ray1.is_parallel(ray2):
             if ray1._is_vertical:
-                x = self._location.x
-                y = ray2.yof(x)
+                x = ray1._location.x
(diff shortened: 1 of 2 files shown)


mozman:
Hi, could you also provide some tests for this edge case?
tests/test_06_math/test_642_construction_line.py

ScottHilton:
Done. I think this should cover it.

mozman:
Thank you!

mozman:
For the future: please also run the tests to check if they pass.

I fixed the problems for this PR.

Thank you!

ScottHilton:
My bad, thanks for all your help. I will remember next time.

PR #131, opened 2020-03-24
Tolerance too low

ScottHilton:
Using ConstructionLine.intersect() I got no intersect where there should be one. 

My error was on the order of 5e-12

Perhaps 1e-10 would be safer though.


ScottHilton:
Example data from running intersection_line_line_2d()

x1 = {float} 616.5
x1_x2 = {float} 0.75
x2 = {float} 615.75
x3 = {float} 616.4025
x3_x4 = {float} 0.0
x4 = {float} 616.4025

x = {float} 616.4025000000014

The input data is clean, yet the output is higher than the tolerance of e-12

mozman:
Thank you, also added `abs_tol` argument to `ConstructionLine.intersect()` and `ConstructionLine.has_intersection()`, because I guess one tolerance value will not fit all user needs.

PR #138, opened 2020-04-15
Query matches unicode string of attribute value

ompugao:
dxf can embed unicode strings in itself.
quotedString only matches ascii strings and that's why this library skips these elements when querying some unicode-string named attributes.


mozman:
How does this improve unicode handling?
Can you show examples for the existing problem?
I will not merge this PR until I see the need for this change.

Issue #141, opened 2020-04-22
Domain error getting angle between vectors


mbway:
Apologies for the bombardment of issues. I'm running into another crash caused by `ezdxf.math.Vector.angle_between()`. The problem is that due to floating point imprecision, the cosine of the angle between two vectors can be outside the domain of the acos function (`-1.0000000000000002`), causing a ValueError to be raised. Below is an example recreating the problem I had with an example solution based off of a stackoverflow question.
```
import math

from ezdxf.math import Vector


def main():
    a = Vector(721.046967113573, 721.0469671135688, 0.0)
    b = Vector(-721.0469671135725, -721.0469671135688, 0.0)

    print(angle_between(a, b))
    print(a.angle_between(b))


def angle_between(a: Vector, b: Vector):
    """ based on https://stackoverflow.com/a/36672201 """
    cos_theta = a.dot(b) / (a.magnitude * b.magnitude)
    abs_cos_theta = math.fabs(cos_theta)
    if abs_cos_theta > 1.0:
        if abs_cos_theta - 1.0 < 1e-5:
            cos_theta = math.modf(cos_theta)[1]
        else:
            raise ValueError('Invalid arguments (vectors not normalized?)')
    return math.acos(cos_theta)


if __name__ == '__main__':
    main()
```
output:
```
3.141592653589793
Traceback (most recent call last):
  File "/home/matthew/bug_report2.py", line 26, in <module>
    main()
  File "/home/matthew/bug_report2.py", line 11, in main
    print(a.angle_between(b))
  File "/home/matthew/anaconda3/envs/physical_map_processing/lib/python3.8/site-packages/ezdxf/math/vector.py", line 477, in angle_between
    return math.acos(self.dot(v2)/self.magnitude/v2.magnitude)
ValueError: math domain error
```

PR #142, opened 2020-04-22
fixed crash caused by floating point inaccuracy in angle_between

mbway:
fixes https://github.com/mozman/ezdxf/issues/141


mozman:
Thank you!

PR #143, opened 2020-04-22
Return skipped entities

mbway:
This pull request adds an interface for determining which entities are skipped during expansion of an Insert entity. The implementation perhaps isn't ideal, but it does limit the amount of change to the implementation of `virtual_block_reference_entities()`.

### DIFF
--- a/src/ezdxf/explode.py
+++ b/src/ezdxf/explode.py
@@ -131,8 +131,7 @@ def virtual_block_reference_entities(block_ref: 'Insert',
         block_ref: Block reference entity (INSERT)
         uniform_scaling_factor: override uniform scaling factor for text entities (TEXT, ATTRIB, MTEXT)  and
                                 HATCH pattern, default is ``max(abs(xscale), abs(yscale),  abs(zscale))``
-        skipped_entity_callback: called whenever the transformation of an entity is not supported and so was
-                                skipped.
+        skipped_entity_callback: called whenever the transformation of an entity is not supported and so was skipped.
 
     .. warning::
 
--- a/src/ezdxf/explode.py
+++ b/src/ezdxf/explode.py
@@ -1,12 +1,13 @@
 # Copyright (c) 2020, Manfred Moitzi
 # License: MIT License
-from typing import TYPE_CHECKING, Iterable, cast, Union, Tuple, Generator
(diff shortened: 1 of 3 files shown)


mozman:
Idea: passing a callback function for skipped entities, this would not mess up the return values.

```python
def virtual_block_reference_entities(block_ref, uniform_scaling_factor=None, notify=None):
    ...
    try:
        copy = entity.copy()
    except DXFTypeError:
        ... 
        notify(entity)
        continue

```

mbway:
yes, that is much nicer. I'll make the change

src/ezdxf/explode.py
```diff
@@ -140,11 +144,15 @@ def virtual_block_reference_entities(block_ref: 'Insert', uniform_scaling_factor
...
+    if skipped_entity_callback is None:
+        def skipped_entity_callback(_entity):
+            pass
 
-    def disassemble(layout):
+    def disassemble(layout) -> Generator['DXFGraphic', None, None]:
         for entity in layout:
             dxftype = entity.dxftype()
             if dxftype == 'ATTDEF':  # do not explode ATTDEF entities
+                skipped_entity_callback(entity)
```
mozman, line 155: This is not necessary because ATTDEF is always replaces by an ATTRIB entity attached to the INSERT entity.

tests/test_04_dxf_high_level_structs/test_414_explode_blockrefs.py
```diff
@@ -155,5 +155,23 @@ def test_05_examine_uniform_scaled_ellipse(doc, msp):
     assert ellipse.dxf.ratio == 0.5
 
 
+def test_06_return_skipped_entities(doc, msp):
```
mozman, line 158: As commented in explode.py, ATTDEF is special and does not really belong to the block content. It is just a template to create attached ATTRIB entities.

src/ezdxf/explode.py
mozman: BTW: The attached ATTRIBs are located in the same coordinate system as the block reference, to examine the attached ATTRIBs just examine `Insert.attribs`

mozman:
Thank you!

PR #145, opened 2020-04-26
ocs_vertices with z && div by zero

vshu3000:
1. in `_transform_2d_ocs_vertices`, the vectors might have 3 components.

2. In two cases, the vector along z axis could have 0 length.

My fix is more of a hack and I fear there are other occasions of normalizing 0-length vectors.  It appears to me that extrusion of (0,0,0) is a common cause of this problem.  Maybe such cases should be systematically detected and fixed.

### DIFF
--- a/src/ezdxf/entities/hatch.py
+++ b/src/ezdxf/entities/hatch.py
@@ -963,7 +963,7 @@ class EdgePath:
 
 
 def _transform_2d_ocs_vertices(ucs, vertices, elevation, extrusion) -> List[Tuple[float, float]]:
-    ocs_vertices = (Vector(x, y, elevation) for x, y in vertices)
+    ocs_vertices = (Vector(v[0], v[1], elevation) for v in vertices)
     return [(v.x, v.y) for v in ucs.ocs_points_to_ocs(ocs_vertices, extrusion=extrusion)]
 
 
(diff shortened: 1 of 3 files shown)


src/ezdxf/entities/hatch.py
```diff
@@ -963,7 +963,7 @@ def export_dxf(self, tagwriter: 'TagWriter') -> None:
 
 
 def _transform_2d_ocs_vertices(ucs, vertices, elevation, extrusion) -> List[Tuple[float, float]]:
-    ocs_vertices = (Vector(x, y, elevation) for x, y in vertices)
```
mozman, line 966: If you get en error here you are doing something wrong, HATCH only supports 2D elements, don't use 3D vertices for hatch boundary paths.

src/ezdxf/entities/insert.py
```diff
@@ -389,11 +389,15 @@ def brcs(self) -> 'BRCS':
         else:
             insert = Vector(insert)
 
+        ez = Vector(self.dxf.extrusion)
```
mozman, line 392: (0, 0, 0) is an invalid extrusion vector, you should not get this value here.

mozman:
I you loaded this DXF files please show me this examples. For everything else: I don't create a fool proof library, this is too much work and makes the code unreadable, learn how the entities work and use sane values. HATCH do not support 3D boundaries and an extrusion vector of (0, 0, 0) is just wrong in both cases an exception is the proper reaction of the library.

PR #146, opened 2020-04-26
Duplicate variable name broke Cython compatibility

DGriffin91:
I was trying to get ezdxf to compile with Cython and was getting this error:

File "acdsdata.py", line 172, in ezdxf.sections.acdsdata.AcDsRecord.__init__
UnboundLocalError: local variable 'tags' referenced before assignment

I was able to get it to compile by changing the duplicate variable name from tags to group.

Thanks for all your efforts on this awesome package!

Just in case you are interested: in my application, after compiling with Cython reading the DXF file is about 3x faster than with pure CPython 3.8.


mozman:
Thank you, using pypy3 can also speed up things for long running applications. Cython optimization is not considered until the v1.0 release.

PR #152, opened 2020-04-29
Ellipse parameters

mbway:
I ran into a problem where ellipses inside blocks were sometimes not rendering correctly. One problem contributing to this is that `angle_to_param()` seemed to be returning incorrect parameters for angles close to the boundaries of the quadrants. Replacing this function with my own implementation fixed that issue.
The unit test ensures that if the ratio is 1.0 then param == angle. Another part to the test is that regardless of other ellipse attributes like the choice of major axis, constructing an ellipse with a parameter calculated from an angle then using `ellipse.(start|end)_point` and getting the angle to that point should yield the original angle. This was not the case with the old implementation.

There were additional issues uncovered by the unit test relating to the fact that the parameter angles of transformed ellipses were not taking direction into account (e.g. `axis = Vector(1, 0, 0)`, `axis.angle_between(Vector(1, -0.1, 0)) == axis.angle_between(Vector(1, 0.1, 0))`), and so the start_point or end_point may not be in the correct location because information is being lost.
To fix this I'm obtaining the anticlockwise angle from the major_axis to the start/end points *about the extrusion vector*.

### DIFF
--- a/src/ezdxf/explode.py
+++ b/src/ezdxf/explode.py
@@ -6,7 +6,7 @@ from typing import TYPE_CHECKING, Iterable, cast, Union, Generator, Callable, Op
 
 from ezdxf.entities import factory
 from ezdxf.lldxf.const import DXFStructureError, DXFTypeError, VERTEXNAMES
-from ezdxf.math import Vector, rytz_axis_construction, normalize_angle, bulge_to_arc, OCS, quadrant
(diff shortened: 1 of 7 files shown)


mbway:
there are still some issues with incorrect results after applying this patch but I'm looking into it. I will add some more unit tests to highlight the issues

mozman:
Thank you very much for your support!

PR #154, opened 2020-05-01
Fix reflections

mbway:
I created this CAD file as reference and tried to match the behaviour when encountering reflections. Before my previous patch, the start and end locations were incorrect, after my ellipse patch the start and end points were correct but sometimes the wrong sides of the ellipses were drawn:
[rotated_and_scaled_arcs.dxf.zip](https://github.com/mozman/ezdxf/files/4563963/rotated_and_scaled_arcs.dxf.zip)

AutoCad:
Screenshot from 2020-05-01 12-29-16
My rendering after the ellipses patch:
ellipse_block_4_negative_scaling_and_rotation
After applying this patch:
ellipse_block_4_negative_scaling_and_rotation

so the ellipses/arcs are correct for all the zscale > 0 cases.
I tried several approaches including negating the extrusion vector, but I found that if a reflected block is exploded in autocad, the extrusion vector is not negated, so instead the ellipse parameters just have to be interpreted differently in the case where a reflection took place. I tried to write logic to perform flips of the start and end parameters when the scale causes a reflection (when Det(scale_matrix) < 0) but I found that just checking whether the start and end have flipped which is bigger is the easiest way to detect that some kind of reflection has occurred.

The text is a bit wrong but I did investigate a little bit to try and fix it, and I don't think the text entities store enough information for them to be reflected outside of a block. if a reflected block is exploded in autocad the text reflection disappears:
Screenshot from 2020-05-01 12-33-43
exploded:
image (2)

I have written unit tests which cover every cell in the table above, but I've disabled the zscale < 0 tests since those fail at the moment. I did attempt to solve these cases but since the point entities weren't even being transformed correctly I think there is an issue elsewhere with the transformation before the ellipse case can be attempted.

### DIFF
--- a/src/ezdxf/explode.py
+++ b/src/ezdxf/explode.py
@@ -206,19 +206,9 @@ def virtual_block_reference_entities(block_ref: 'Insert',
 
         dxftype = entity.dxftype()
 
+        original_ellipse: Optional[Ellipse] = None
         if has_non_uniform_scaling and dxftype == 'ELLIPSE':
-            # transform start- and end point before main transformation
-            ellipse = cast('Ellipse', entity)
-            open_ellipse = not math.isclose(
-                normalize_angle(ellipse.dxf.start_param),
-                normalize_angle(ellipse.dxf.end_param),
-            )
-            if open_ellipse:
-                # transform start- and end point
-                start_param = ellipse.dxf.start_param
-                end_param = ellipse.dxf.end_param
-                start_point, end_point = brcs.points_to_wcs(ellipse.vertices((start_param, end_param)))
-            minor_axis = brcs.direction_to_wcs(ellipse.minor_axis)
+            original_ellipse = entity.copy()
 
         # Basic transformation from BRCS to WCS
         try:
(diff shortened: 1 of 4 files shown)


mbway:
I've found that there are some cases where ellipses still don't render correctly with zscale > 0. I think I have neglected the case where the untransformed ellipse is drawn clockwise

mozman:
Great job, this works better than I ever thought, don't bother to much with text.

mozman:
Hi!

What is worse than no documentation? Wrong documentation (comments).

Reflection about all three axis (-1, -1, -1) was handled as uniform scaling, switching the flag `has_non_uniform_scaling` explicit to `True` creates this result (left=blocks, right=exploded):

image

Still not perfect, but may be a hint what the problem was.

mbway:
I've been trying to find out what cases my fix doesn't cover. I've found one at least but I know there are more. Unfortunately I have to move on temporarily but will hopefully fix the remaining issues at some point.
take an arc with major axis (1, 0, 0) drawn from 0.0001 to pi/2, scale by (-1, 1, 1). The transformed ellipse is now from tau-0.0001 to 3pi/2 so start and end must be flipped, this is fine.
however
take an arc drawn from 0 to pi/2, scale by (-1, 1, 1) and this time the ellipse is from 0 to 3pi/2 so start and end won't be flipped in this case, although they should be.

also I've realised that although my rendering checks that the correct side of the ellipse is drawn, the unit test does not, it only checks the start and end points. In addition to `Ellipse.vertices()` it would be useful to have something like `Ellipse.params(num: int)` to sample parameters evenly between the start and end parameter. I have a function for this in my own code but I'm not sure if it is correct because for my purposes I don't care about the direction so this generates samples going anticlockwise regardless of the ellipse given to it:
```python
        if draw_angles is None:
            start, end = 0, 2 * pi
        else:
            start, end = draw_angles
            if start >= end:
                end += 2 * pi
        params = np.linspace(start, end, num=num)
```
with a function like `Ellipse.params()`, the unit test could also make sure the vertex half way along the ellipse is in the correct place as well.

mozman:
The ARC problem is in the OCS/WCS transformation, as I wrote this code axis scaling wasn't planned, just simple UCS transformation.

mozman:
I think for OCS entities like ARC and TEXT and non uniform scaling the extrusion vector has to be recalculated from the new plan and not transformed as direction vector, because it is not sure the transformed direction is a normal vector to the new plane.

mozman:
Since I have already traveled halfway down the rabbit hole, it is probably the right time to implement a proper transformation interface `DXFEntity.transform(t: Matrix44)`. That would also be a good time to get numpy on board, unfortunately there is no precompiled wheel available for pypy3 on Windows.

Therefore numpy support can only be an optional feature, which adds complexity. I will stay with the already implemented `Matrix44()` class for now.

mozman:
New transform interface:

above: block references; below: exploded by ezdxf 'transform' branch

image

ELLIPSE, ARC and TEXT looks very good, MTEXT reflections need some work.

PR #166, opened 2020-06-20
Drawing

mbway:
an add-on for visualising dxf documents and either saving them to files (e.g. png, svg etc) or viewing them interactively (see cad_viewer.py)

currently, PyQt5 and Matplotlib backends are provided, but other backends should not be too difficult to implement as necessary (for example, a 3D backend with OpenGL could be a possible addition)


mbway:
a demonstration of this addon: https://youtu.be/F_JKh4WcTTs

PR #178, opened 2020-07-13
Adding support for Linux and fixing windows GUI pop-up problem

George-Jiang:
The original one can be only working under Windows. Errors occur when working under Linux.
I added support for Linux and further fixed the Windows pop-up problem.
I slightly adjusted the process by making a temp_folder under the working directory then deleting it, rather than using the os temp_path which caused an error in linux.

### DIFF
--- a/src/ezdxf/addons/odafc.py
+++ b/src/ezdxf/addons/odafc.py
@@ -5,6 +5,9 @@ from typing import Optional
 import logging
 import subprocess
 import os
+import sys
+import uuid
+from shutil import rmtree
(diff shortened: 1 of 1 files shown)


mozman:
Sorry, I don't have a VirtualBox installation anymore to test this changes, I just use the headless windows subsystem for linux.

George-Jiang:
result = subprocess.Popen(args=cmd, stdout=subprocess.PIPE, stderr=None, startupinfo=startupinfo)
result.wait()  #this line need to be added in the previous commit

PR #188, opened 2020-07-14
In documentation of the "draw" addon, set "layout" object?

mherzog01:
I am a total newbie to GitHub, so I hope I'm doing this right.  Also, I'm also new to ezdxf, so be kind regarding the following question/suggestion.

In the documentation of the Draw addon, the code references a "layout" object, but the "layout" object is not initialized, so the code fails.  My pull request attempts to correct this.

Thank you for this awesome package.  I hope this helps!


mozman:
Thank you, I  will fix this.

PR #192, opened 2020-07-19
odafc addon improvements

mbway:
I saw some activity from pull requests to get the oda file converter addon working on Linux. I have already solved the GUI problem for linux and thought I should share my findings.

Since ezdxf seems to prefer fewer dependencies I didn't use the xvfbwrapper python package but instead wrote a minimal xvfb interface to allow the ODAFileConverter to run inside a dummy X display.

While I was at it I did some refactoring, added some more checks for files already existing and used `shutil.which()` which avoids hard coding paths for linux and hopefully OSX (but Windows might still require the hard coded path). Also TemporaryDirectory is a good cross platform way of obtaining a temporary directory rather than hard coding anything.

### DIFF
--- a/src/ezdxf/addons/odafc.py
+++ b/src/ezdxf/addons/odafc.py
@@ -1,19 +1,22 @@
 # Copyright (c) 2020, Manfred Moitzi
 # License: MIT License
 # Created: 2020-04-01
-from typing import Optional
 import logging
-import subprocess
 import os
+import platform
+import shutil
+import subprocess
+import tempfile
+import time
+from contextlib import contextmanager
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


mbway:
setting the environment variable `QT_QPA_PLATFORM` to `offscreen` (the solution from https://github.com/mozman/ezdxf/pull/178) didn't work for me on Ubuntu 18.04. I get the error:
```
This plugin does not support propagateSizeHints()
QPainter::begin: Paint device returned engine == 0, type: 2
Segmentation fault (core dumped)
```
which is why the xvfb solution is a better option. It should work on any Linux distro using X11.

PR #193, opened 2020-07-19
Miscellaneous Drawing Changes

mbway:
I've collected together a few changes I wanted to make with the drawing add-on.

The commit messages explain the functionality, but an overview is:
- dwg support in cad_viewer
- click to cycle through currently selected entities in draw order in cad_viewer (now that many entities are drawn as paths, their collision rectangles obscure any entities behind them making this feature more necessary)
- shift + clicking on a layer checkbox will toggle every layer in cad_viewer
- while the cad file is being drawn, a 'loading' overlay is shown in cad_viewer
- when a layout is first drawn, the checked state of each layer checkbox is determined by the initial layer visibility
- a few other minor things

I don't have time to fix the visibility issue right now which is outlined in the new unit tests in 9159ce8 which currently fail. I'm not sure what I should do with this. Should I leave them, skip them with `pytest.skip()`, or remove them from this pull request and post an issue instead?
The original drawing add-on as of eba112815f8b65a99e91fb0d1ec7d70cb80757df had the correct visibility behaviour so there has been a regression at some point. The tests are a executable form of the 'visibility.dxf' example file which I used to informally verify the behaviour originally. the [notes/drawing.md](https://github.com/mozman/ezdxf/blob/master/docs/notes/drawing.md#layers-and-draw-order) document explains the expected visibility behaviour.


mozman:
Hi!

OK, I (don't) understand the problem: the visibility in respect of layer state of an INSERT depends on the content of the INSERT and not on the layer of the INSERT. I don't use AutoCAD very much, and this behavior is not what I am expected. 

Solution: Excluding INSERT from visibility check

... some testing later ... OK it is more complex - Layer `0` is special

I will modify this tests to match the previous behavior - visibility depends only the INSERT entity,
it is simple, I understand this behavior and it is different to AutoCAD.

mozman:
Maybe this special behavior of layer '0' is also important for other DXF attributes and attribute resolving has to consider this fact?

Ideas: 

- If an entity is inside of a BLOCK (INSERT) and resides on layer '0', resolve entity layer as INSERT.dxf.layer, 
  this could also avoid some special checks in the resolving process.
- new method: RenderContext.resolve_layer(entity)
- renaming `current_block` into `current_block_reference` (done)
- renaming `is_block_context` into `inside_block_reference` (done)
- condition check: RenderContext.inside_block_reference

PR #199, opened 2020-07-29
Added caret support in TEXT entity

mbway:
I realised that my last pull request left some things unfinished because TEXT entities should also support caret encoding.
It turns out that they do, but handle them slightly differently.
- non-printable characters are replaced with `?` rather than `▯`
- fewer characters are supported in TEXT, for example `\t` is replaced with `?`
- `\n` does have an effect, but seems to not be intended behaviour to allow this so I just remove `\n`

There is also a difference in that MText strings can end with a `^` and still be valid and rendered correctly, however a `Text` ending with `^` will be invalid and AutoCAD will refuse to load the file. I added a check for this during export but I'm not sure where the best place to put this check is, so you might want to move it.

AutoCAD rendering
image (4)
ezdxf rendering
caret

### DIFF
--- a/src/ezdxf/addons/drawing/backend.py
+++ b/src/ezdxf/addons/drawing/backend.py
@@ -7,6 +7,7 @@ from typing import Optional, Tuple, TYPE_CHECKING, Iterable
 from ezdxf.addons.drawing.properties import Properties
 from ezdxf.addons.drawing.type_hints import Color
 from ezdxf.entities import DXFGraphic
+from ezdxf.entities.mtext import replace_non_printable_characters
 from ezdxf.math import Vector, Matrix44
 from ezdxf.render.path import Path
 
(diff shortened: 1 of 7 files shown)


mozman:
Hi!

Thank you for all your support!

In the past I often raised exceptions even when a small fix is possible, because I thought it is better to show such errors, but now I tend to fix such problems silently. (like '^' at the end of TEXT)

In this case I will remove the '^' char at export/audit silently and raise an error in the `add_text()` factory function to show the usage error, but the original design as simple DXF interface does not catch `Text.dxf.text = '^' `, but at least the DXF export is not invalid.

mozman:
For now I will not add a validity check to the factory functions, this is just code duplication, and using the audit process for every new entity will slow down the whole package and adds the requirement of a valid DXF Document for the factory interface, which I just removed for the `VirtualLayout()` class. First goal is, not to export invalid DXF files, the package user can get more information by starting the audit process manually, an extra topic to this problem in the documentation is required - which very few package-user will ever read.

mozman:
Started adding input validators for simple DXF attributes located in the `DXFNamespace.__setter__()`, but this will also prevents loading DXF files with minor DXF flaws, like an color index of -1, and fix them.

There are 2 ways:

- AutoCAD (the unfriendly CAD) does't not load a DXF file including color index -1
- BricsCAD changes color index -1 to 256 (BYLAYER)

I think I will go the BricsCAD way, although I would like to process only valid DXF files, it is a fact that there are many invalid DXF files in the wild, caused by the lack of a good DXF reference, thx Au7ode5k.

PR #229, opened 2020-09-09
Qsave improvements

mbway:
a few small things changed with the quick save utility:
- I am not sure if there would ever be a use for any rect other than (0, 0, 1, 1) since the image could just be loaded and manipulated afterwards if the user truly only wanted part of the image to be filled with the CAD drawing. If you think it would be useful then I can revert that change
- `transparent=True` sets the axis color to fully transparent and `facecolor=...` sets the figure color. If the output format doesn't support transparency then the output will just become opaque so I don't think there is a need to expose a parameter for `transparent`
    - if neither `transparent` or `facecolor` is used then the output will be opaque even if the cad/axis background color is semi-transparent  
    - if only transparent is used then the background is fully transparent
    - if only facecolor is used then the alpha will be twice what is should be (because both the figure and axes have the same alpha and are overlaid on top of each other)
    - so both facecolor and transparent need to be set to get the correct alpha value in the output image
- by default, the matplotlib backend will probably be something like `qt5Agg` or `tkAgg` which have support for calling `plt.show()` (opening a GUI window), however a warning will be raised if a figure is created on any thread other than the main thread. Since `plt.show()` will never be called for this function, it is a good idea to use a non-interactive backend which can only write to files (and restore the backend afterwards)

### DIFF
--- a/src/ezdxf/addons/drawing/matplotlib.py
+++ b/src/ezdxf/addons/drawing/matplotlib.py
@@ -188,9 +188,7 @@ def _get_path_patch_data(path):
 def qsave(layout: 'Layout', filename: str, *,
           bg: Optional[Color] = None,
           fg: Optional[Color] = None,
-          rect: Sequence[float] = (0, 0, 1, 1),
           dpi: int = 300,
-          transparent: bool = False,
           ) -> None:
     """ Quick and simplified render export by matplotlib.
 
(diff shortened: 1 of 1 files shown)


mbway:
by the way, although threads are uncommon in python programs due to the GIL, one use case is rendering and saving an image in the background of a PyQt application, which does use threads.

src/ezdxf/addons/drawing/matplotlib.py
```diff
@@ -188,40 +188,51 @@ def _get_path_patch_data(path):
 def qsave(layout: 'Layout', filename: str, *,
           bg: Optional[Color] = None,
           fg: Optional[Color] = None,
-          rect: Sequence[float] = (0, 0, 1, 1),
```
mozman, line 191: I had also the feeling that this argument is not necessary.

mozman:
Thank you!

I don't think I  fully understand the transparency logic, but a first test showed the expected results!

```Python
doc = ezdxf.readfile('fillings.dxf')
msp = doc.modelspace()
# black background:
matplotlib.qsave(msp,  "bg_black.png", bg='#000000')
# white background
matplotlib.qsave(msp,  "bg_white.png", bg='#ffffff')
# transparent background, black foreground color (ACI=7)
matplotlib.qsave(msp,  "bg_transparent.png", bg='#ffffff00')
```

PR #234, opened 2020-09-19
separated out _get_text_path for easier optimization

DGriffin91:
I'm trying to improve draw speeds with the pyqt backend. Text seems to be one of the bottlenecks. I've tripled the performance of  `draw_text_entity_2d()`  by using the built-in python LRU cache on text path creation. (tested with my production cad files) 

This pull request makes it easier for me to patch in my caching system. 

I figured you would probably not want to add the LRU cache to master at this point. In case you are curious about my implementation, I currently have this in cad_viewer.py:
```
@lru_cache(maxsize=2048)
def _get_text_rect(font: qg.QFont, text: str) -> qc.QRectF:

    split_text = text.split()
    if len(split_text) == 1:  # No spaces
        return _get_text_path(font, text).boundingRect()

    width = 0
    height = 0
    space_width = _get_text_path(font, " ").boundingRect().width()
    for word in split_text:
        new_size = _get_text_path(font, word).boundingRect()
        width += new_size.width() + space_width
        height = max(height, new_size.height())
    width -= space_width
    return qc.QRectF(0, 0, width, height)


ezdxf.addons.drawing.pyqt._get_text_rect = _get_text_rect


@lru_cache(maxsize=2048)
def _get_text_path(font: qg.QFont, text: str) -> qc.QRectF:
    path = qg.QPainterPath()
    path.addText(0, 0, font, text)
    return path


ezdxf.addons.drawing.pyqt._get_text_path = _get_text_path
```

### DIFF
--- a/src/ezdxf/addons/drawing/pyqt.py
+++ b/src/ezdxf/addons/drawing/pyqt.py
@@ -153,8 +153,7 @@ class PyQtBackend(Backend):
         scale = cap_height / self._font_measurements.cap_height
         transform = Matrix44.scale(scale, -scale, 0) @ transform
 
-        path = qg.QPainterPath()
-        path.addText(0, 0, self._font, text)
+        path = _get_text_path(self._font, text)
         path = _matrix_to_qtransform(transform).map(path)
         item = self.scene.addPath(path, self._no_line,
                                   self._get_color(properties.color))
@@ -203,10 +202,14 @@ def _matrix_to_qtransform(matrix: Matrix44) -> qg.QTransform:
     return qg.QTransform(*matrix.get_2d_transformation())
 
 
-def _get_text_rect(font: qg.QFont, text: str) -> qc.QRectF:
+def _get_text_path(font: qg.QFont, text: str) -> qc.QRectF:
     path = qg.QPainterPath()
     path.addText(0, 0, font, text)
-    return path.boundingRect()
+    return path
+
+
+def _get_text_rect(font: qg.QFont, text: str) -> qc.QRectF:
+    return _get_text_path(font, text).boundingRect()
 
 
 def _get_font_measurements(font: qg.QFont) -> FontMeasurements:


mozman:
@mbway Could you check this PR?

mozman:
@DGriffin91 The current version supports only one default font, but support for multiple fonts is planned, this could have an impact on your caching strategy.

DGriffin91:
@mozman I think my caching strategy could still be somewhat effective with multiple fonts. Also, the majority of the actual CAD drawings I deal with on a daily basis only have one font face, and they have that face at the same height 90% of the time. 

I see how this optimization probably doesn't work too well in a general application, but in my specific application it should still help. That said I'm obviously very open to better ways of making this faster to redraw. I'm already planing on separating out the drawing into layers, etc.. and trying to only update what is needed. 

Also, I'm hoping to use Cython or something else similar to help. (I've already had success with Cython and ezdxf in other applications)

mbway:
caching text paths is definitely a good idea but I don't think LRU is the best eviction strategy for this. If your cache has size N but your cad file has even just N+1 unique strings, then re-drawing will cause a 100% cache miss rate since entities are always drawn in the same order. The cache probably has the most potential for speeding up redraws rather than within a single draw since there probably won't be massive amounts of identical text. However the backend implementation is currently a bit naive in that the path has to be calculated once to get just the rect (for laying out the text), then again when actually drawing, so a cache definitely helps here as well.

I think it would be better to have an unlimited cache owned by the backend (rather than a global like `@lru_cache(maxsize=None)` would be). That way if the user is worried about RAM constraints, they can simply tell the backend not to cache anything, but if they are not worried then the backend can cache text paths between different uses and even across different CAD files.

This should work with multiple fonts as well. I have come up with an implementation where the backend could keep a dictionary like `font_family->text renderer` (or similar) in the future so the paths for each family can be cached independently.

mbway:
also, in terms of speeding up the Qt backend generally. I've found that if the only thing that changes is visibility of some entities, it is much better to simply identify those entities in the scene and set their visibility accordingly rather than redrawing. I'm able to toggle layer visibility almost instantly even when dealing with cad files which take >10 seconds to render initially.

The cad_viewer doesn't do this for simplicity reasons, but after drawing, you could iterate over the scene items and create a lookup of `entity_id -> scene item` so that you can quickly identify the scene items you want to set visible/invisible.

you could do something similar if you want to redraw certain layers that have changed in some way, just remove the scene items of the old layers and replace them. It might be tricky to retain the correct draw order however since layers are independent of draw order. If no entities are added or removed you could record the z value for the old scene items before removing them.

DGriffin91:
@mbway Thanks for looking into this! As well as providing some insight on avoiding re-draw. For editing the drawing, it seems like marking what needs to be updated, and only redrawing those entities would work well. Though when doing an operation like moving a large number of entities it wouldn't be real time. It seems like there would have to be a preview showing the movement by just transforming entities in the scene, then redrawing after the operation is complete. Optimizing other common operations would be even more complex.

I think your caching strategy in #235 makes a lot of sense. On my test files, the performance on initial draw in #235 is just a tad slower than what I had, but the redraw is a bit faster. I think it's just because I was caching individual words and there are a lot of repeated words in my production files.

mbway:
looking at your implementation, it looks like you would still have to render the entire path once since `_get_text_path` would be called with the entire string but the rect calculation might be faster from the per-word caching.

by the way there is a bug in your rect implementation because `space_width` will always be 0.0. You might not have noticed this if you aren't using word wrapping. If you aren't using word wrapping then you could override the PyQtBackend and just return 0.0 for `get_text_line_width` and everything else should work. That way you wouldn't have to calculate the bounding rects at all.

Another thing you could try if you do need word wrapping is to approximate the bounding rect rather than generating the word paths to obtain the width. Maybe something like this? I tested it on the text stress test I included in #235 but for that file the rect calculation is pretty insignificant so this wasn't noticably faster. Also the results aren't identical to using `boundingRect()` so this is probably slightly less accurate but it seems sufficient to render the `examples_dxf/text.dxf` accurately.

```python
class FastTextRenderer(TextRenderer):
    def __init__(self, font: qg.QFont, use_cache: bool):
        super().__init__(font, use_cache)
        _metrics = qg.QFontMetricsF(font)
        self._width_lookup = {chr(i): _metrics.horizontalAdvance(chr(i)) for i in range(256)}

    def get_text_rect(self, text: str) -> qc.QRectF:
        width = sum(self._width_lookup[c] for c in text)
        return qc.QRectF(0, 0, width, self._font_measurements.cap_height)

    def _get_font_measurements(self) -> FontMeasurements:
        upper_x = self.get_text_path('X').boundingRect()
        lower_x = self.get_text_path('x').boundingRect()
        lower_p = self.get_text_path('p').boundingRect()
        return FontMeasurements(
            baseline=lower_x.bottom(),
            cap_top=upper_x.top(),
            x_top=lower_x.top(),
            bottom=lower_p.bottom(),
        )

```

DGriffin91:
@mbway I do have word wrapping in my examples, and did notice it was off. I was planning on taking a look at that but hadn't yet. The spacing is a bit off from what I have in autocad even using master. Everything looks like it is a bit left of where it should be or something. I haven't looked into that more yet. I'm assuming it has to do with not using the same font or something.

I had thought of doing a look up with each character but it would seem that it would then not be taking into account kerning. Though, for bounding boxes maybe that doesn't matter too much. 

On my cad files, your FastTextRenderer is about 10% faster for initial loading and about the same for redraw.

It feels like at this point it's probably good enough to use what you have in #235 and try to avoid redraw when possible. 

When profiling I did notice that in `pyqt.py` in `draw_text()` that `path = _matrix_to_qtransform(transform).map(path)` is fairly slow. Not sure if there is a good way of speeding that up that isn't unnecessarily complicated. Just for fun I tried caching it, but it wasn't really faster, I think making the hash of the matrix, etc. was too slow. I even tried caching pretty much all of `draw_text()` just to see and it still only made a small difference since `scene.addPath()` is what's really slow. Which comes back to avoiding redraw altogether.

mozman:
replaced by PR #235

PR #261, opened 2020-11-10
Update lwpolyline.rst

OkanPinar:
Adding formula for radius calculation. It can be critical for creation curve with these data.


mozman:
Or you look at the "see also" section and click at the link [Bulge Related Function](https://ezdxf.mozman.at/docs/math.html#bulge-related-functions).

PR #304, opened 2021-01-11
Math pickling

### LINKED ISSUE #303: pickle.dump() raises error with Vec3 (Vector) at ezdxf v0.15
Pickling variables include Vec3 (Vector) raises TypeError at ezdxf v0.15. Is it a bug?
I'm using v0.14 as a provisional measure.

```
test.py, line 7, in <module>
    pickle.dump(test_data, f)
  File "stringsource", line 2, in ezdxf.acc.vector.Vec3.__reduce_cython__
TypeError: no default __reduce__ due to non-trivial __cinit__
```

Code below is a simple example to reproduce this error.
```
import pickle
from ezdxf.math import Vector

test_name = "test.pickle"
test_data = Vector(1, 1, 1)
with open(test_name, "wb") as f:
    pickle.dump(test_data, f)

with open(test_name, "rb") as f:
    result = pickle.load(f)

print(result)
```

mbway:
added support for pickling the cython math classes since I knew it would be easy to do. Fixes #303

### DIFF
--- a/src/ezdxf/acc/matrix44.pyx
+++ b/src/ezdxf/acc/matrix44.pyx
@@ -45,6 +45,9 @@ cdef class Matrix44:
             raise ValueError("invalid argument count: 4 row vectors or "
                              "iterable of 16 numbers")
 
+    def __reduce__(self):
+        return Matrix44, (tuple(self),)
+
     def __getitem__(self, tuple index: Tuple[int, int]) -> float:
         cdef int row = index[0]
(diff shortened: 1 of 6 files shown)


mbway:
by the way, I noticed that `vec2` cannot be initialised with `vec2()` because the pure python implementation doesn't support it, so I removed `vec2()` from the valid initialisations in the cython documentation

PR #320, opened 2021-01-21
improved TEXT alignment calculation

mbway:
this fixes #319. The behaviour is reverse-engineered from what I see from rendering the text stress test. I checked and the stress test covers all possible alignment values and since it renders correctly I assume the logic is correct.

@mozman do you know about any other edge cases that might not be covered? for example is `align_point` a valid attribute of ATTRIB? Also do you agree with the logic in `_get_wcs_insert`?

### DIFF
--- a/src/ezdxf/addons/drawing/text.py
+++ b/src/ezdxf/addons/drawing/text.py
@@ -44,6 +44,25 @@ AnyText = Union[Text, MText, Attrib, AttDef]
 # multiple of cap_height between the baseline of the previous line and the baseline of the next line
 DEFAULT_LINE_SPACING = 5 / 3
 
+DXF_TEXT_ALIGNMENT_TO_ALIGNMENT: Dict[str, Alignment] = {
+    'LEFT': (HAlignment.LEFT, VAlignment.BASELINE),
+    'CENTER': (HAlignment.CENTER, VAlignment.BASELINE),
+    'RIGHT': (HAlignment.RIGHT, VAlignment.BASELINE),
+    'ALIGNED': (HAlignment.RIGHT, VAlignment.BASELINE),
+    'MIDDLE': (HAlignment.CENTER, VAlignment.LOWER_CASE_CENTER),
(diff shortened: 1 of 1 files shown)


mozman:
The insert location is `Text.dxf.insert` if alignment is (and only if) `BASELINE/LEFT` else `Text.dxf.align`. `FIT` and `ALIGNED` fits the text between `insert` and `align`, `FIT` keeps the height `ALIGN` adjusts the height. Maybe `FIT` and `ALIGN` should be treated like `MIDDLE`. It's not easy to create `FIT` and `ALIGNED` text entities in CAD applications, you have to use the `_TEXT` command, therefore I guess they are not really common.

image

See also: [Text.set_pos()](https://github.com/mozman/ezdxf/blob/b6c1794bb8928b59ab092f595c82fa17d4d4de2b/src/ezdxf/entities/text.py#L185).


Example file with FIT and ALIGNED text: [dropbox link](https://www.dropbox.com/s/u6xw0hvczhhdgcq/FIT_ALIGNED_TEXT.dxf?dl=0)

mbway:
there are fit and aligned examples in the stress test. I did notice that with the new layout logic they now appear right-aligned. I assume that if the font measurements were perfect then right-aligned at the align point is exactly the same as left-aligned at the insert point. Still, it might be worth using the insert point for these alignments as left-alignment is more natural? or finding the midpoint between the insert and align points and doing center alignment as you suggest is another good idea.

I'll also change the logic to use the insert pos if LEFT-BASELINE aligned if that follows the spec more accurately.

mozman:
This image shows red the INSERT points shifted by (1, 1), only the `BASELINE` alignment does change:

image

There are the specs ;-)

image

mozman:
The `ALIGNED` and the `FIT` alignment ignore the stored text rotation angle and calculates the rotation from the `insert` to the `align_point`. The text is fitted between these two points, and `MIDDLE` is really like `MIDDLE_CENTER`:
image

PR #339, opened 2021-02-02
quadratic to cubic bezier utility

mbway:
in my application I only want to have to support one type of bezier curve, so I needed a utility to convert quadratic curves in order to support the new `Command.CURVE3_TO`.

I might have not put the function in the right module. I can move it if you want.

### DIFF
--- a/src/ezdxf/math/__init__.py
+++ b/src/ezdxf/math/__init__.py
@@ -38,7 +38,7 @@ from .bspline import (
 )
 from .bezier import Bezier
 from .bezier4p import (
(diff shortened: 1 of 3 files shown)


mbway:
by the way, in what situations do quadratic beziers show up? can dxf represent them or was this addition only to support fonts?

mozman:
Location is OK! The quadratic Bézier curves are the preferred curve type used in TTF fonts, at least in the fonts I have used to test.

mozman:
DXF can't represent them, but I always wanted to implement them, because they are supported by the most backends, except Cairo. But until the `text2path` add-on, there was just no need for that.

PR #345, opened 2021-02-05
fixing __geo_interface__ for polygon

mamofejo:
For polygons without holes the geo_interface was missing the outer list.


mozman:
What does this mean?

mozman:
thx for the bug report

PR #401, opened 2021-04-18
added support for viewport clipping.

Rahulghuge94:
hello sir, i have just created function for clipping viewport. i have tested it and it is working. Hope will be useful for everyone.

### DIFF
--- a/src/ezdxf/entities/viewport.py
+++ b/src/ezdxf/entities/viewport.py
@@ -458,3 +458,15 @@ class Viewport(DXFGraphic):
             handle = self.dxf.clipping_boundary_handle
             return handle != '0'
         return False
+    
+    def clip_viewport(self,path_entity):
+        '''Clips viewport for given path entity.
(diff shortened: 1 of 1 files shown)


mozman:
Thank you.

Tests are required. Examples for the different supported DXF entities as clipping path would be nice. Have you tested the saved DXF files for all supported DXF versions R12 - R2018 with AutoCAD or TrueView?

mozman:
If you replace an existing path entity with a new one, should't you remove the reactor handle from the previous path object?

mozman:
You should also check if the new path entity is alive!

mozman:
How do 3D entities behave as 2D clipping path? To be be clear, every DXF entity can be an 3D entity by using the OCS.

mozman:
If you provide the user an interface to use any DXF entity, you also have to check if the entity resides in the same layout and even if the entity is from the same document. Users will try to abuse your interface, see the user always as your enemy. I have learned this lesson the hard way.

mozman:
Insecure implementation and no testing

PR #447, opened 2021-05-27
improved ezdxf color palette accuracy

mbway:
I noticed an issue where some entities which were visible in AutoCAD were almost invisible in ezdxf because of slight differences in the color palette (specifically color 250). So I decided to extract the exact AutoCAD 2020 color palette myself rather than looking online.
I created a dxf file with this script:
```python
import ezdxf

def main():
    doc = ezdxf.new()
    modelspace = doc.modelspace()
    paper_space = doc.layouts.new('paper space')

    for layout in (modelspace, paper_space):
        for i in range(260):
            height = 1
            points = [(0, 0), (2, 0), (0, height), (2, 1)]
            rows = 30
            this_i = i % rows
            this_y = this_i * height + (this_i-1) * 0.1
            this_x = -(i // rows) * 6
            points = [(x + this_x, y + this_y) for x, y in points]
            layout.add_solid(points, {'color': i})
            layout.add_text(str(i), {'insert': (this_x + 2, this_y), 'height': 1})
    doc.saveas('/tmp/colors.dxf')


if __name__ == '__main__':
    main()
```
then used this script to pick colors of a screenshot of that dxf file taken in autocad:
(i did the whole process twice to make sure there were no errors)
```python
import matplotlib.pyplot as plt
import sys

if __name__ == '__main__':
    fig, ax = plt.subplots()
    fig.tight_layout()
    ax.set_aspect('equal', 'datalim')
    img = plt.imread(sys.argv[1])
    ax.imshow(img)

    def on_click(event):
        if event.inaxes is not None:
            r, g, b = img[int(event.ydata), int(event.xdata), :3] * 255
            print('0x{:02x}{:02x}{:02x},'.format(int(r), int(g), int(b)))

    fig.canvas.mpl_connect('button_press_event', on_click)
    plt.show()
```
here are the screenshots for comparison
autocad:
model_space
paper_space

old ezdxf:
image (47)
image (48)


mbway:
one thing that ezdxf currently doesn't handle is that some colors change slightly between the modelspace and paperspace (250-256). I'm not sure how best to handle these cases

PR #480, opened 2021-06-21
Include py.typed file in package to comply with PEP 561

johnthagen:
See https://www.python.org/dev/peps/pep-0561/

`mypy` and other type checkers use the presence of a `py.typed` file in the root of the package to signal that the package has inline type annotations to be used for type checking. Including this file allows users of this library to properly type check their usage of this package.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,6 @@
 include NEWS.md README.md LICENSE
 include requirements.txt
+include src/ezdxf/py.typed
 recursive-include src/ezdxf/acc *.hpp *.cpp *.h *.c *.pxd *.pyx
 recursive-include tests *.py
 recursive-include tests/data *
--- a/setup.py
+++ b/setup.py
@@ -95,6 +95,7 @@ setup(
         'tools/font_face_cache.json',
         'tools/font_measurement_cache.json',
         'resources/*.png',
+        'py.typed',
     ]},
     entry_points={
         'console_scripts': [
--- a/src/ezdxf/py.typed
+++ b/src/ezdxf/py.typed
@@ -1,0 +1,2 @@
+# Instruct type checkers to look for inline type annotations in this package.
+# See PEP 561.


mozman:
You seem to have much more "typing" experience than I do.

I only used the included type checking capabilities from PyCharm. Running `mypy` reveals "some" more errors (> 2000 ;)

Perhaps you have a few tips to improve compatibility with `mypy` or to configure some very strict rules away from` mypy`.

My first question would be: How to improve usage of `Unions`, see `mypy` errors in https://github.com/mozman/ezdxf/blob/master/src/ezdxf/tools/binarydata.py
That was also the search in which I found your posting: https://github.com/python/mypy/issues/1693
```Text
py.exe -m mypy ezdxf\tools\binarydata.py
ezdxf\tools\binarydata.py:420: error: Incompatible return value type (got "Union[float, Sequence[float]]", expected "Sequence[float]")
ezdxf\tools\binarydata.py:426: error: Incompatible return value type (got "Union[float, Sequence[float]]", expected "float")
ezdxf\tools\binarydata.py:429: error: Incompatible return value type (got "Union[int, Sequence[int]]", expected "int")
ezdxf\tools\binarydata.py:433: error: Argument 1 to "range" has incompatible type "Union[int, Sequence[int]]"; expected "SupportsIndex"
ezdxf\tools\binarydata.py:440: error: Unsupported operand types for * ("Sequence[int]" and "int")
ezdxf\tools\binarydata.py:440: note: Left operand is of type "Union[int, Sequence[int]]"
ezdxf\tools\binarydata.py:460: error: Incompatible return value type (got "Tuple[Union[int, Sequence[int]], str, str]", expected "Tuple[int, str
...
```

I opened a new issue for such hints and tips: https://github.com/mozman/ezdxf/issues/481

johnthagen:
@mozman 

> I only used the included type checking capabilities from PyCharm.

I use it primarily for PyCharm as well, but I've recently begun to incorporate `mypy` type checking as well so that CI can typing errors as well.

> Running mypy reveals "some" more errors (> 2000 ;)

This is a very common occurrence for a large code base that hasn't yet used `mypy`. Given how dynamic Python is and that PyCharm is much more loose about what it considers a "type error", the first go at using `mypy` takes a bit of work. I have found that overall, using `mypy` is a net benefit, especially if you are already putting in the time to correctly annotate 

A few suggestions:

- Start small. You can point `mypy` at only a subset of your project, and work up from there
- Be liberal with `# type: ignore` at first. This allows you to suppress certain typing failures while you work out  a complete first pass. You can then come back to those later, but at least moving forward new code will be type checked.
- Use `typing.cast`, `isinstance()`, and `is not None` checks as needed. This allows you to communicate to `mypy` when a certain type is narrower than a particular `Union` / `Optional`
- Add type checking to CI. I have an example that uses `tox` and GitHub Actions on my python-blueprint project: https://github.com/johnthagen/python-blueprint/blob/master/tox.ini#L45

johnthagen:
You can also disable or enable certain mypy checks globally using a configuration file: https://mypy.readthedocs.io/en/stable/config_file.html#

You could use this to disable certain checks during your first pass, get the suite passing, and then selectively re-enabling certain checks one and at a time.

PR #498, opened 2021-07-20
added basic OLE2FRAME drawing support

mbway:
for me, OLE2FRAME entities are are in the same category as Viewport: I don't really care what they contain but I would like to know that they are there.

Currently OLE2FRAME entities are skipped but I think it would be better for the frontend to handle these entities when identified. In the future someone could expand on the `draw_ole_frame_entity` method to draw the entity properly if they want to.

I know that this implementation isn't ideal as it reads dxf tags directly, but it's quite simple and adds the minimum support for the entity needed to draw it as a rectangle. If this isn't sufficient I can make some improvements with your guidance.

### DIFF
--- a/src/ezdxf/addons/drawing/frontend.py
+++ b/src/ezdxf/addons/drawing/frontend.py
@@ -1,12 +1,13 @@
 # Copyright (c) 2020-2021, Matthew Broadway
 # License: MIT License
 import math
-from typing import Iterable, cast, Union, List, Dict, Callable
+from typing import Iterable, cast, Union, List, Dict, Callable, Tuple
 from ezdxf.lldxf import const
 from ezdxf.addons.drawing.backend import Backend
 from ezdxf.addons.drawing.properties import (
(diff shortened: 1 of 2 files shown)


mozman:
I am sure you know what OLE data is: https://en.wikipedia.org/wiki/Object_Linking_and_Embedding
Useless proprietary data that should be erased from the world! :)

mbway:
totally agree! It's one of the reasons I started the ezdxf drawing addon: to hopefully free more people from having to use proprietary tools. I don't know how successful that has been overall but the company I work for has almost entirely removed it's reliance on AutoCAD in large part thanks to ezdxf :)

mbway:
I just realised there is a bug. Sorry about that. I shouldn't have wrapped the results of `get_first_value` in `Vec3` until I know that they aren't `None`. I don't think I can append a pull request. Should I open another one?

mozman:
I will fix this.

PR #513, opened 2021-07-30
3DFACE: per edge visibility [WIP]

mbway:
The individual edges of a 3DFACE entity can be visible or hidden. Previously the `RenderContext` would cast `entity.dxf.invisible` to a bool to determine whether a 3DFACE is visible or not. For 3DFACE this field is a bit mask specifying the invisibility of each edge rather than the visibility of the entire entity, so if any edge was not visible the entire entity would be skipped.

With this pull request the entity is considered visible if any of it's edges are visible and the frontend considers two cases:
- the case where every edge is visible: draw as a path like before
- the case where one or more edges are invisible: draw the visible edges as individual lines

I did consider using `move_to` and always drawing as a path but that may complicate the backend as currently `move_to` commands are not expected in the middle of paths.

There are some examples whose visibility is still incorrect. I think the fix for this would involve changing how 3DFACE is handled. At the moment the rectangle and triangle cases are treated distinctly but I'm not sure if this is a good idea, because I think the more general case is that one or more points are identical. This can take many forms such as the 'lines' in the examples below. Also in the unit test a newly created Face3d is a 'triangle' because all it's vertices are (0, 0, 0). In terms of drawing the entity I think it would be fine for the frontend to unconditionally draw 4 edges for every 3DFACE even if some lines have 0 length, and the backend should filter those out. Since ezdxf isn't just used for drawing I'm not sure if any other components find the quad/tri distinction useful which is why I'm asking rather than just changing it.

[3dface.dxf.zip](https://github.com/mozman/ezdxf/files/6907963/3dface.dxf.zip)

AutoCAD:
3dface

Ezdxf:
image

### DIFF
--- a/src/ezdxf/addons/drawing/frontend.py
+++ b/src/ezdxf/addons/drawing/frontend.py
@@ -399,17 +399,24 @@ class Frontend:
     def draw_solid_entity(
         self, entity: DXFGraphic, properties: Properties
     ) -> None:
-        assert isinstance(
-            entity, (Solid, Face3d)
-        ), "API error, requires a SOLID, TRACE or 3DFACE entity"
-        dxf, dxftype = entity.dxf, entity.dxftype()
-        points = entity.wcs_vertices()
-        if dxftype == "3DFACE":
-            self.out.draw_path(from_vertices(points, close=True), properties)
-        else:
+        if isinstance(entity, Face3d):
+            points = entity.wcs_vertices(close=True)
+            edge_visibility = entity.get_edges_visibility()
+            if all(edge_visibility):
+                self.out.draw_path(from_vertices(points), properties)
+            else:
+                assert len(points) - 1 == len(edge_visibility)
+                for a, b, visible in zip(points, points[1:], edge_visibility):
+                    if visible:
+                        self.out.draw_line(a, b, properties)
+
+        elif isinstance(entity, Solid):
             # set solid fill type for SOLID and TRACE
             properties.filling = Filling()
-            self.out.draw_filled_polygon(points, properties)
+            self.out.draw_filled_polygon(entity.wcs_vertices(close=False), properties)
+
+        else:
+            raise TypeError("API error, requires a SOLID, TRACE or 3DFACE entity")
 
     def draw_hatch_entity(
         self,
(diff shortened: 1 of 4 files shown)


mozman:
Great job!

In my opinion the entities SOLID, TRACE and 3DFACE could always return 4 vertices from the methods `vertices()` and `wcs_vertices()` if this leads to a more consistent handling. The `make_path()` function for these entities does not care about edge visibility and always creates a closed path (0-1-2-(3-)0). A `line_to` command for two coincident points will be ignored: a "triangle" (0-1-2-2), creates the path (0-1-2-0)

I understand the the 3DFACE entity as a 3D patch and don't know if the strange examples (points and lines) are important in real world applications. Did you see this configuration in real world examples? Some DXF configurations can only be created by writing the DXF file, but not by CAD applications. 

A real world scenario I think of is dividing entities, which can happen across a 3DFACE, but I assume most CAD applications will create two closed  triangles from a quadrilateral if necessary and not two triangles with an invisible edge at the cutting location.

mozman:
BricsCAD shows the same results as AutoCAD and this is the 3D View:

image

mbway:
> Did you see this configuration in real world examples

I think I've seen triangles before. The lines were something I added to see whether they rendered correctly with the current quad/tri handling. The example dxf was created in autocad with no post-processing so all the entities in the example could appear in real CAD files.

Quite often I have seen 3DFACE used to create 2D shapes for some reason. Most recently there was one used to create a door where one of the edges was hidden, so it created a `U` like shape which the door arc connects to. I found these entities because after ezdxf filtered them out completely it was obvious that there was something missing. I'm not really sure why 3DFACE was favoured over polyline or rectangle but I just have to support what I find.

mbway:
in `export_entity()` there is an if statement:
```python
        if not self.dxf.hasattr("vtx3"):
            self.dxf.vtx3 = self.dxf.vtx2
```
so in the case of a triangle is the vtx3 sometimes omitted (in my examples they are not but other cad programs might)? in which case reading from it would default to `(0, 0, 0)` wouldn't it? if so then this still needs to be handled properly rather than just saying for example:
```python
        vertices = [dxf.vtx0, dxf.vtx1, dxf.vtx2, dxf.vtx3]
```
or was that check introduced as a precaution but you haven't seen that case in real files?

PR #542, opened 2021-09-23
improved configuration of the drawing addon

mbway:
following the discussion https://github.com/mozman/ezdxf/discussions/534 I have refactored the drawing addon configuration so that the backend is accessed through a strictly controlled interface rather than accessing and having the frontend and render context modify public attributes in-place.

The `Configuration` class is read only (frozen) so that the flow of data is easier to track. No longer can other parts of the system mutate values all over the place.

I also took the opportunity to replace some of the integer arguments with enums so that their possible values are clearer. I was not entirely sure about the purpose of some of the parameters, for example `lineweight_scale` vs `linetype_scale`. It seemed that `linetype_scale` is a hack where 1 => interpret line styles and 0 => draw solid lines but I'm not sure. I removed this parameter in favour of incorporating the 'draw solid' option into the `LinePolicy` enum.

There were a few tricky cases where it seems that different values are set for different backends. I think the overall configuration struct should be backend independent, and if a backend needs extra information to be able to follow those instructions then they should be given that information directly. For example the Qt backend has an `extra_lineweight_scaling` which allows it to scale line weights to match what the user would expect, rather than having the configuration object itself be influenced by which backend you are using. Similarly, the `min_lineweight` was being set in `qsave` so I changed that parameter to have a value (None) where the backends themselves will choose an appropriate value.

The configuration class could be split into parameters which are relevant to the frontend only and those which are to configure the line rendering etc but for now I have just collected all the parameters into a single class.

### DIFF
--- a/src/ezdxf/addons/drawing/backend.py
+++ b/src/ezdxf/addons/drawing/backend.py
@@ -1,6 +1,6 @@
 # Copyright (c) 2020-2021, Matthew Broadway
 # License: MIT License
-from abc import ABC, abstractmethod
+from abc import ABC, abstractmethod, ABCMeta
 from typing import (
     Optional,
     Tuple,
@@ -12,6 +12,7 @@ from typing import (
     Union,
 )
 
+from ezdxf.addons.drawing.config import Configuration
 from ezdxf.addons.drawing.properties import Properties
 from ezdxf.addons.drawing.type_hints import Color
 from ezdxf.entities import DXFGraphic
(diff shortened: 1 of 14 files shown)


mbway:
I don't understand the mypy failure. the error doesn't seem related to the configuration changes

mozman:
What is the advantage of the `Configuration.defaults()` function over default values in the `dataclass` definition?

mbway:
The advantage isn't huge so I'm fine either way, but my thinking was that it allows old code to continue working the same even if breaking changes are made, because the user can choose to explicitly provide a value for every setting, and if new settings are created in the future that will throw an error rather than run but potentially produce a different output to what you were expecting.

And then people that don't care about that can use `Configuration.defaults()` or `Configuration.defaults().with_changes(...)`

It also gives the opportunity to provide multiple sets of defaults, for example `Configuration.high_quality_defaults()`?

mbway:
Just read through the changes again. Quite a lot changed so there may be a few mistakes like this:
I forgot to use `PyQtBackend._extra_lineweight_scaling`. It should be used here I think:
```python
        px = properties.lineweight / 0.3527 * self.config.lineweight_scaling * self._extra_lineweight_scaling
```
Shall I make another pull request or can you fix that?

PR #556, opened 2021-10-07
Use of entity handle as id for SVG export

bdaniere:


mozman:
Some ideas if you need this feature for your project: 

1. The creation of the `gid` should look like this:
```Python
from typing import Optional
from uuid import uuid4
from ezdxf.entities import DXFGraphic

def get_gid(entity: Optional[DXFGraphic]) -> str:
    if entity is None:
        return ""

    handle = entity.dxf.handle
    if handle is not None:
        return handle

    # entity has no handle
    suffix = ""
    if entity.has_source_block_reference:
        if not entity.source_block_reference.is_virtual:
            suffix += "_" + entity.uuid.hex
            entity = entity.source_block_reference
        else:
            while entity.has_source_block_reference:
                suffix += "_" + entity.uuid.hex
                entity = entity.source_block_reference

    assert entity is not None, "this should not happen"

    # handle is None
    entity = entity.origin_of_copy
    if entity is not None:  # doesn't have to have an origin -> virtual entity
        handle = entity.dxf.handle

    if handle is None:
        # virtual entity without a handle or handle is None
        handle = uuid4().hex
    return handle + suffix
```

2. No need to change `Frontend` or `Properties`, the backend has access to the currently processed entity, example for `draw_point()`:
```Python
def draw_point(self, pos: Vec3, properties: Properties):
    gid = get_gid(self.current_entity)
    ...

```
3. Write your own backend based on `ezdxf.addons.drawing.matplotlib.py`
4. Write tests
5. Run tests

avinash-218:
> I don't know if I will merge this code. Is this helpful for anyone else?

Yess helpful for me

Issue #584, opened 2021-12-07
Add support to release linux aarch64 wheels


odidev:
I’m trying to build Linux AArch64 wheels but the repository [mozman/manylinux-wheel-builder](https://github.com/mozman/manylinux-wheel-builder) does not have aarch64 support. 

@mozman , Could you please let me know your plan for adding AArch64 support in the same. I can help with this.

mozman:
I am not the creator of the `manylinux-wheel-builder`, I just use it and I have no idea how to add this feature.

PR #586, opened 2021-12-14
Add Linux AArch64 wheel build support

odidev:
Added linux aarch64 wheel build support.
Related to https://github.com/mozman/ezdxf/issues/584, @mozman Could you please review this PR?

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -9,15 +9,30 @@ jobs:
   # Build and deploy manylinux wheel
   Linux-build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform: [x86_64, aarch64]
     env:
       TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
       TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
     steps:
       - uses: actions/checkout@v2
+      - name: Set up QEMU
+        if: ${{ matrix.platform == 'aarch64' }}
+        uses: docker/setup-qemu-action@v1
       - name: Build and upload manylinux wheels
+        if: ${{ matrix.platform == 'x86_64' }}
         uses:  mozman/manylinux-wheel-builder@master
         with:
           python-versions: "3.7, 3.8, 3.9, 3.10"
+        # if true then github actions won't stop even if build for this job fails
+        continue-on-error: true
+      - name: Build and upload manylinux aarch64 wheels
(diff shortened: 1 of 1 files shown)


mozman:
Is there no PyPI upload? Or how can I add the PyPI upload? The PyPI upload is the only purpose of this workflow!

mozman:
added upload to PyPI ...

mozman:
Why do you use tag v0.3.4-manylinux2014_aarch64 instead of the latest version v0.4.2-manylinux2014_aarch64?

odidev:
The latest version v0.4.2-manylinux2014_aarch64 was released recently. Before that I was using v0.4.0-manylinux2014_aarch64 which was causing error in auditwheel command due to platform mismatch. So, I used older version. However I just checked with the latest version and it also looks to be working fine.

mozman:
A test run with v0.4.2-manylinux2014_aarch64 and deployment to [TestPyPI](https://test.pypi.org/project/ezdxf/0.17.2b3/#files) was successful.

Can you do a test install with any of these packages?

odidev:
It's working.

mozman:
Thanks for your feedback, but I just changed the workflow to use `pypa/cibuildwheel@v2.3.1` to support more platforms, the first release `v0.17.2b4` is running now (ARM64 build on QEMU takes a looooooong time) and can be tested soon: https://pypi.org/project/ezdxf/0.17.2b4/

odidev:
I installed the wheel and ran tests on both x86 and aarch64. On running the tests on both the platforms,  I'm getting some assertion errors. Please have a look at the logs for the same: 

**Test logs on x86_64:** [ezdxf_test_x86.txt](https://github.com/mozman/ezdxf/files/7734664/ezdxf_test_x86.txt)
**Test logs on aarch64:**[ezdxf_test_arm64.txt](https://github.com/mozman/ezdxf/files/7734665/ezdxf_test_arm64.txt)

mozman:
The x86 tests (205, issue_589) which are failing were added today and can only run with the latest version fro github - the wheels were build yesterday!

Same for the arm64 platform but there are additional floating point impressions error in: 621, 630a, 708b, 906 which I try to solve.

To run the test suite you ALWAYS have to install the latest version from github or you have to use the tests included in the source code distribution for the wheel you use, the tests are not included in the wheels itself. For non-beta releases the "stable" branch is also a way to go.

mozman:
Install current version from github: https://ezdxf.mozman.at/docs/setup.html#installation-from-github

PR #740, opened 2022-09-03
Fix typo in browse-command config example for gedit

musicinmybrain:
The result still doesn’t seem to work for me (on Fedora Linux 36) for reasons that seem to be related to quoting and/or argument splitting. I haven’t taken time to figure out exactly what is happening.

Still, this PR fixes the typo in the command path.


mozman:
Thank you!

PR #820, opened 2023-01-18
for make_path() bulges respect segments except hatch

nikolarobottesla:
1. converter.py: get segments from kwargs in 2/3 uses of add_2d_polyline (hatch doesn't have kwargs)
2. tools.py: add_2d_polyline(): bulge_to(): added segmentation of angles and loop
3. path.rst: Path Factory Function description added BULGE


nikolarobottesla:
@mozman atting you because I can't add reviewers.

mozman (approved):

PR #858, opened 2023-03-30
tools.py/add_2d_polyline()/bulge_to() ensure shortest arc fix

nikolarobottesla:
Prior to this bug fix bulge angles that were > pi between the start and end point. For example -100deg to 200deg would be sampled in the wrong direction around the circle.

Instead of implementing this in tools.py it could maybe go in bulge.py/bulge_to_arc() but that would effect more functions and maybe it would mess up negative bulges.

I tested both negative and postive bulges with the proposed implmentation and got expected results.

### DIFF
--- a/src/ezdxf/path/tools.py
+++ b/src/ezdxf/path/tools.py
@@ -655,6 +655,13 @@ def add_2d_polyline(
         # each cubic_bezier adds 3 segments, need 1 minimum
         num_bez = math.ceil(segments / 3)
         center, start_angle, end_angle, radius = bulge_to_arc(p1, p2, bulge)
+        # sample angles of the shortest arc between start and end angles
+        central_angle = end_angle - start_angle
+        if abs(central_angle) > math.pi:  # if difference between angles greater than pi
+            if central_angle > 0:
+                start_angle = start_angle + math.tau  # shift the start point by a full rotation
+            else:
+                end_angle = end_angle + math.tau  # shift the end point by a full rotation
         angles = list(linspace(start_angle, end_angle, num_bez + 1))
         curves = []
         for i in range(num_bez):


mozman:
Is this really tested?

A bulge value >1 creates arc with angles > pi:

image

mozman:
I tested your code and it seems to do no harm,  but show me an error which this PR fixes.

mozman:
This is the order of Bézier curves for a bulge value of 2:
image

This is the order of Bézier curves for a bulge value of -2:
image

I don't see the problem.

nikolarobottesla:
-1 to 1 fix enabled
image

nikolarobottesla:
My use case is importing a dxf and converting the lwpoline into a point list to do computational geometry. I need the higher resolution on the bulges that I get from my last PR.  Is there a better way to get a point list than path.make_path()?

mozman:
OK, I understand the problem: for bulge values of 1 and greater the arc crosses the 0/360 degree border which yields total crap - I will fix this.

Increasing the segment count is not the way to solve your specific problem. Use `Path.flattening()` to convert a path into a sequence of vertices, the distance argument defines the max. sagitta,  the max. distance between the curve and the approximated polyline.

```Python
import ezdxf
from ezdxf import path

doc = ezdxf.new()
msp = doc.modelspace()
pline = msp.add_lwpolyline(([(0, 0), (1, 0, 0, 0, 0.9), (3, 0), (4, 0)]))
p = path.make_path(pline)
msp.add_lwpolyline(p.flattening(0.1), dxfattribs={"color": 1})
doc.saveas("pline.dxf")

```
image

The segment argument is meant to increase the precision of the circular arc approximation by Bèzier curves, which cannot represent circular arcs with 100% precision. The more segments, the more precise the approximation.

mozman:
This should be fixed now: bulge > 1 with 12 segments == 3 Bèzier curves

image

Feel free to open a new issue if you find another error and thank you for the error report.

nikolarobottesla:
Thanks for fixing the bulge thing and the tip about path flattening.

PR #880, opened 2023-05-13
Configurable cad view

mbway:
Prompted by https://github.com/mozman/ezdxf/discussions/878 I had a go at creating a simple example of using the Qt drawing functionality to select an element but found that the code was not organised to make this very easy.

This PR splits out the cad widget functionality from CADViewer into a separate CADWidget class which is more suitable for embedding into other applications or injecting custom functionality (such as selecting elements).

There are also some other minor refactorings:
- renamed some `Cad*` classes to `CAD*` to match the convention elsewhere such as `DXFEntity` instead of `DxfEntity`
- fixed what I think is a typo in the beginner docs
- renamed the `element_selected` signal to `element_hovered` which is more specific

### DIFF
--- a/docs/source/usage_for_beginners.rst
+++ b/docs/source/usage_for_beginners.rst
@@ -262,7 +262,7 @@ attribute :code:`doc.layers`.
 
 You can create your own layers::
 
-    my_layer = doc.layer.add("MyLayer")
+    my_layer = doc.layers.add("MyLayer")
 
 The layer object also controls the visibility of entities which references this
 layer, the on/off state of the layer is unfortunately stored as positive or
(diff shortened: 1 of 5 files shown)


Gashmen:
Hello, honored ones. I need to create my own interface based on the Cad Widget, when the user selects some properties from the widget sheet will change the Drawing itself and I want users to immediately see the changes on the CadWidget, rather than opening the dxf. I also want to provide for the possibility of moving elements in the connection CadWidget - change coordinates - passing coordinates to doc.modelspace().

image


@mbway answered many questions, would appreciate help or direction to the right path.

Issue #949, opened 2023-10-30
pymupdf integration is broken


planted-sam:
**Describe the bug**
When following the example for PDF generation with the drawing addon [(specifically the example for PDF creation)](https://github.com/mozman/ezdxf/blob/stable/examples/addons/drawing/pdf_export.py), I keep running into an error where PDF generation fails.

Specifically, the `get_pdf_bytes` method call is where the function breaks down, with the trace going back to `PyMuPdfRenderBackend.set_background` in `ezdxf/addons/drawing/pymupdf.py`. That method fails because when it tries to run `shape.drawRect` it throws `AttributeError: 'Shape' object has no attribute 'drawRect'`.

Tracing where the shape object comes from, it looks like it uses the fitz module from pymupdf, and looking at the most recent pymupdf docs,[ it seems like the Shape class doesn't have a `drawRect` method, but rather a `draw_rect` method. ](https://pymupdf.readthedocs.io/en/latest/shape.html)It makes me think that maybe that function name changed in pymupdf and the change hasn't been reflected in this codebase.

Could this code be updated to match the new class definition in pymupdf? Or is there a old version of pymupdf that I should install specifically instead of the latest one to get this to work?

**To Reproduce**
Run the example at https://github.com/mozman/ezdxf/blob/stable/examples/addons/drawing/pdf_export.py with the latest version of pymupdf installed (1.23.5), and see the error that pops up when `get_pdf_bytes` is called.

**Expected behavior**
The `set_background` method of the PyMuPDFRenderBackend should complete without error.

planted-sam:
Update: I forked the library and made the minor edits to the function calls, then tested again with my forked/edited version of the library and it worked with no issue. I've opened a pull request with the changes, I'm happy to continue the discussion on that PR or here. Thanks!

mozman:
Thank you!

PR #950, opened 2023-10-30
Update references to pymupdf's Shape class to use new function names

planted-sam:
## Details
Referencing #949

This updates the functions called on pymupdf's Shape class to match the current function names for pymupdf.

Tested locally with a sample PDF generation that worked with these changes, and failed without them. The test PDF I used was a relatively simple test case (only line and circle entities on a dead simple layout), but I'm including it here for reference anyway.

Let me know how/if I can run other tests or do anything else to show that this change works

## Sample output from testing
[sample.pdf](https://github.com/mozman/ezdxf/files/13208969/sample.pdf)


mozman:
Thank you, I fixed this issue manually.

mbway:
the minimum version should be set to 1.18.14 [source](https://pymupdf.readthedocs.io/en/latest/znames.html#deprecated) to ensure that the snake case names are available

PR #960, opened 2023-11-10
Added save convert function

nbowisdar:
Sometimes the value can have a look:
    1.000000E 20
    -1.000000E 20
    and when we put it into a float() we get a ValueError
    
   I couldn't put this function into  repair.py because of "partially initialized module" error.
   You can access filesfor testing here 
   https://drive.google.com/file/d/1l_ls3R3QkfRlvjXXo1JumB-25FAnCjVR/view?usp=sharing

### DIFF
--- a/src/ezdxf/lldxf/tagger.py
+++ b/src/ezdxf/lldxf/tagger.py
@@ -17,6 +17,7 @@ from .types import (
     BINARY_DATA,
 )
 from .const import DXFStructureError
+from .utils import value_to_float
 from ezdxf.tools.codepage import toencoding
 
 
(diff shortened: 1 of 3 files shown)


mozman:
Hi!

Thank you for your contribution.

Most of these files can be opened by the `recover` module, "open_bug_3.dxf" is the only file which can't be opened by the `recover` module. I will improve the `recover` module to read these male formed floating point values.

mozman:
The file "draw_bug_3.dxf" contains errors which the `recover` module fixes by deleting the entity.

These are some possible causes for these errors:

- document has entities without handles (invalid for DXF R13+)
- $HANDSEED is not the next usable handle
- entity gets an already used handle
- entity overwrites existing entity or will be overwritten by an entity  loaded afterwards

Just the start of the audit report:
```
auditing file: draw_bug_3.dxf
58 issues fixed.

   1. Issue [12] fixed.
   Removed entity POLYLINE(#90) with a conflicting handle and without a database entry.

   2. Issue [12] fixed.
   Removed entity POLYLINE(#726) with a conflicting handle and without a database entry.

   3. Issue [12] fixed.
   Removed entity POLYLINE(#729) with a conflicting handle and without a database entry.

   4. Issue [12] fixed.
   Removed entity POLYLINE(#72c) with a conflicting handle and without a database entry.

   5. Issue [12] fixed.
   Removed entity POLYLINE(#72f) with a conflicting handle and without a database entry.

   6. Issue [12] fixed.
   Removed entity POLYLINE(#732) with a conflicting handle and without a database entry.

   7. Issue [12] fixed.
   Removed entity POLYLINE(#7e8) with a conflicting handle and without a database entry.

...
```

mozman:
I found the issue of "draw_bug_3.dxf": 
Every POLYLINE entity has the same handle as its first VERTEX entity. That means the VERTEX entity replaces the POLYLINE entity in the entity database at the loading stage and this cannot be fixed in `ezdxf`.

image

PR #965, opened 2023-11-17
fixed a typo in ezdxf.path.rect()

neiljackson1984:
There seems to be a missing minus sign in the `ezdxf.path.rect()` function.  As a result, the final vertex is the upper-right corner, not the expected lower-right corner, and you end up with a right triangle instead of the expected rectangle.   I have fixed the problem by adding the missing minus sign.


mozman:
Thank you!

PR #975, opened 2023-11-26
Convert examples/copydxf.py from CRLF to LF lines

musicinmybrain:
Match other source files, which all have UNIX-style LF line terminations.

The following DXF files are also text with CRLF lines and could also be converted, if you like. I’m happy to add them to this PR.

```
integration_tests/data/bin_dxf_r2000.dxf
integration_tests/data/bin_dxf_r12.dxf
integration_tests/data/bin_dxf_r14.dxf
integration_tests/data/bin_dxf_r13.dxf
tests/test_01_dxf_entities/houses_of_parliament_georeferenced.dxf
examples_dxf/text_negative_extrusion.dxf
examples_dxf/text.dxf
```


musicinmybrain:
> DXF files created on Windows by AutoCAD and BricsCAD uses CRLF, it's a wonder that AutoCAD accepts just LF, but converting them to LF is not required,

Okay, that makes sense.

Any interest in converting the Python source file in this PR?

mozman:
I'm on a Windows computer, so there should be more than one file with CRLF line endings.

I worked with PyCharm until last month and then switched to VSCode. I don't recognize these different line endings of files and Git also seems to ignore them by default. Therefore I see no reason at all to care about different line endings.

PR #985, opened 2023-12-08
Fix ResourceWarning about unclosed file on POSIX

danofsteel32:
The warning:

```
odafc.py:145: ResourceWarning: unclosed file <_io.BufferedReader name=5>
  _execute_odafc(args)
```

On POSIX using `subprocess.run` is simpler to use and reason about since it just blocks until the process returns.


mozman:
Thank you!

PR #994, opened 2023-12-15
Correct attribute name: landing_gap_size

seanpk:
ref: `src/ezdxf/render/mleader.py`


mozman:
Thanks, but it's too late for that. I don't break existing code for that.

seanpk:
I don't understand, I think maybe you misunderstood what the PR was for...?

 The documentation does not match the code. How would updating the documentation, which is what this PR is for, break any existing code?

mozman:
I misunderstood that.

mozman:
Wrong target branch.

PR #1000, opened 2023-12-29
Improve draw command

mbway:
as part of fixing the svg and mupdf backends I wanted a way to *use* the backends. I have abstracted the file output logic from the ezdxf draw command so now all the backends can be used with ezdxf draw:

- **matplotlib**: bitmap and vector
- **qt**: bitmap and vector (svg) (the output is created without opening a GUI)
- **mupdf**: bitmap and vector
- **custom_svg**: vector (svg)

use with `ezdxf draw <dxf_path> -o <out_path> --backend <backend>`

this may be useful for development (to compare outputs between backends) and also may be useful for users because one backend may give the result that they want while another might not. Also one backend may be faster than another.

### DIFF
--- a/src/ezdxf/addons/drawing/config.py
+++ b/src/ezdxf/addons/drawing/config.py
@@ -122,8 +122,8 @@ class BackgroundPolicy(Enum):
 
     Attributes:
         DEFAULT: as resolved by the :class:`Frontend` class
-        WHITE: white background
-        BLACK: black background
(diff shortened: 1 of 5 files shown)


mbway:
I'm aware that the handling of dpi is all over the place. Not sure how to bring them in line so that the same value produces roughly the same result regardless of the backend. It seems that for mupdf a larger dpi results in a smaller image for example

PR #1001, opened 2023-12-29
Drawing image support 2

mbway:
this adds support for rendering images in the muPDF backend.
I also improved the missing image text rendering a bit so the text should always have a reasonable size and is now centered where it should be. The output of mupdf is looking very nice :)

one issue is that mupdf seems to have issues with transparency gradients in images, but I don't think there is much ezdxf can do about that.
ezdxf + mupdf (to png)
images
reference (autocad)
image

### DIFF
--- a/src/ezdxf/addons/drawing/frontend.py
+++ b/src/ezdxf/addons/drawing/frontend.py
@@ -66,7 +66,7 @@ from ezdxf.entities.attrib import BaseAttrib
 from ezdxf.entities.polygon import DXFPolygon
 from ezdxf.entities.boundary_paths import AbstractBoundaryPath
 from ezdxf.layouts import Layout
-from ezdxf.math import Vec2, Vec3, OCS, NULLVEC
+from ezdxf.math import Vec2, Vec3, OCS, NULLVEC, Matrix44
 from ezdxf.path import (
     Path,
     make_path,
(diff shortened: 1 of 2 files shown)


src/ezdxf/addons/drawing/frontend.py
```diff
@@ -711,12 +711,26 @@ def draw_image_entity(self, entity: DXFGraphic, properties: Properties) -> None:
                 )
 
             elif show_filename_if_missing:
```
mbway, line 713: I'm not sure whether the filename takes on the transformation of the image (eg rotating, mirroring, stretching etc) as my example file does not have examples of that. This implementation will apply the transformation and will ignore clipping as well

src/ezdxf/addons/drawing/pymupdf.py
```diff
@@ -358,7 +362,38 @@ def draw_filled_polygon(
...
+                    expand=True,
+                    fillcolor=(0, 0, 0, 0),
+                )
+            image = np.asarray(pil_image)
+            height, width, depth = image.shape
+
+        pixmap = fitz.Pixmap(fitz.Colorspace(fitz.CS_RGB), width, height, bytes(image.data), True)
+        # TODO: could improve by caching and re-using xrefs. If a document contains many
+        #  identical images redundant copies will be stored for each one
+        self.page.insert_image(r, keep_proportion=False, pixmap=pixmap)
```
mbway, line 396: insert_image is quite limited in what it supports.
- It can only insert images that are axis-aligned (there is a rotation argument but must be a multiple of 90 degrees)
- the rect argument controls translation and scaling
- but with `keep_proportion=False` it can handle non-uniform scaling (but only axis-aligned)
- if the image has reflection or rotation I resample the image, for example (where white would be transparent):
image282

the alternative approach would be to create a mini pdf document + page, insert the image then use `self.page.show_pdf_page()` which would avoid the need for resampling, but might be more tricky to get working

PR #1004, opened 2024-01-03
Added additional dir for fonts on mac

ru4ert:
do i need to update a version?

### DIFF
--- a/src/ezdxf/fonts/font_manager.py
+++ b/src/ezdxf/fonts/font_manager.py
@@ -33,7 +33,10 @@ LINUX_FONT_DIRS = [
     "~/.local/share/fonts",
     "~/.local/share/texmf/fonts",
 ]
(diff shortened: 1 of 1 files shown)


mozman:
Thank you!

I don't understand your question.

When you add a new search directory, you have to [rebuild](https://ezdxf.mozman.at/docs/tools/fonts.html#rebuilding-the-font-cache) the font-cache: `ezdxf --fonts`
This is also required when you install the next version of `ezdxf` which includes this change.

ru4ert:
No, you misunderstand me ^^

I don't found contribution guidelines. 
In many projects you have to make a version tag, in java you have to update the **pom.xml**, in JS/TS you update the **package.json**.

Every project has a different deployment pipeline or so and i didn't know if there is smth similar in python for this project. ^^

mozman:
I do it all myself, no need to worry.

PR #1011, opened 2024-01-20
Fix the infinite loop in `dxf2code.py`

guyi2000:
- Fix infinite loop in `merge()` function


guyi2000:
The current implementation involves merging `self.code` into itself, resulting in an infinite loop. The correct approach is to merge `code.code`.

PR #1093, opened 2024-05-26
New vector property: is_axial

toporivskiy:
Added a function that determines whether the vector is directed along one of the axes

### DIFF
--- a/src/ezdxf/math/_vector.py
+++ b/src/ezdxf/math/_vector.py
@@ -292,6 +292,15 @@ class Vec3:
             and abs(self._z) <= ABS_TOL
         )
 
+    @property
+    def is_axial(self) -> bool:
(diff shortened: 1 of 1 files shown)


mozman:
Thank you for your contribution, but I will not merge this, because I see no need for these methods.

There is much more to do when you add things to the core library especially to classes with Python and Cython implementations. Tests for new methods are mandatory and the docs should also be updated.

mozman:
A helper function is much more convenient than extending 4 (2x Python and 2x Cython) classes:

```Python
def is_axial(vec: Iterable[float], abs_tol=1e-12) -> bool:
    return any(abs(a) < abs_tol for a in vec)
```

This function works  with `Vec2` , `Vec3` and any iterable of floats.

PR #1111, opened 2024-06-20
enable pickle entity

GoatWang:
Enable pickling entity. Since the circular reference of entity.dxf and dxf._entity, causing the fail pickling. To address this, _entity attribute of entity.dxf will be removed before `pickle.dump` and will be rewired when `pickle.load`. Also, test function is added to valid the feature.

### DIFF
--- a/tests/test_01_dxf_entities/test_110_dxfentity.py
+++ b/tests/test_01_dxf_entities/test_110_dxfentity.py
@@ -185,6 +185,23 @@ def test_delete_missing_source_block_reference_without_exception():
     e.del_source_block_reference()
     assert True is True
 
+def test_pickle():
+    import pickle
+    from io import StringIO
+    from ezdxf.lldxf.tagwriter import TagWriter
+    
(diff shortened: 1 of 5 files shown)


mozman:
I don't see any need to support the `pickle` protocol, and I'm not sure if this won't create more problems than it solves.

The DXF structures have many and often deep linkage, just testing with `DXFEntity` is not sufficient. Test your code with `HATCH` or `MLEADER` entities.

tests/test_01_dxf_entities/test_110_dxfentity.py
mozman, line 1: To clarify my problem with this change:

When you pickle an entity from a loaded DXF file, the entity contains a reference to this document and the whole DXF document will be included in the data dump.

GoatWang: Thanks for the clarification, I will continue to learn this issue, try to verify it and keep the changes on my fork for now.

PR #1116, opened 2024-06-26
Allows setting/resetting `entity.rgb` to `None`

kkkkkom:
Hi, @mozman , could you please review this minor merge request?

I think we should allow `entity.rgb` to be set/reset to `None`.

Many entity rgbs are initially `None` by default, and this change will allow users/developers to temporarily set rgb and reset it back without causing an error.

pytest has passed clean locally.

### DIFF
--- a/src/ezdxf/entities/dxfgfx.py
+++ b/src/ezdxf/entities/dxfgfx.py
@@ -238,8 +238,22 @@ class DXFGraphic(DXFEntity):
 
     @rgb.setter
     def rgb(self, rgb: clr.RGB) -> None:
-        """Set RGB true color as (r, g , b) tuple e.g. (12, 34, 56)."""
-        self.dxf.set("true_color", clr.rgb2int(rgb))
+        """
+        Set RGB true color as (r, g , b) tuple e.g. (12, 34, 56).
+        Also allows setting/resetting rgb to None.
+        """
+        if rgb is None:
+            del self.rgb
+        else:
+            self.dxf.set("true_color", clr.rgb2int(rgb))
+
+    @rgb.deleter
+    def rgb(self) -> None:
+        """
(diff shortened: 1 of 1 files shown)


mozman (changes requested):
This is not the correct way.

If you rally need this feature implement the `@rgb.deleter` method and check if the attrib `true_color`  exists otherwise a `DXFAttributeError` will be raised.

kkkkkom:
> This is not the correct way.
> 
> If you rally need this feature implement the `@rgb.deleter` method and check if the attrib `true_color` exists otherwise a `DXFAttributeError` will be raised.

Thanks for reviewing, I'll update accordingly

kkkkkom:
Hi, @mozman , could you please review again?

src/ezdxf/entities/dxfgfx.py
```diff
@@ -238,8 +238,22 @@ def rgb(self) -> Optional[clr.RGB]:
 
     @rgb.setter
     def rgb(self, rgb: clr.RGB) -> None:
-        """Set RGB true color as (r, g , b) tuple e.g. (12, 34, 56)."""
-        self.dxf.set("true_color", clr.rgb2int(rgb))
+        """
+        Set RGB true color as (r, g , b) tuple e.g. (12, 34, 56).
+        Also allows setting/resetting rgb to None.
+        """
+        if rgb is None:
```
mozman, line 245: This case should raise a TypeError, None is not a valid input or revert everything because the previous version raised the TypeError automatically.

mozman:
@kkkkkom  thanks for the idea, to speed up the process I added all the necessary changes myself

kkkkkom:
> @kkkkkom thanks for the idea, to speed up the process I added all the necessary changes myself

Thanks!

PR #1132, opened 2024-07-27
Fix units.unit_name to work on Python 3.11 or later

### LINKED ISSUE #1131: `ezdxf.units.unit_name` always returns `unitless` in some environment
Thank you for maintaining this awesome package!

**Describe the bug**
`ezdxf.units.unit_name` should return the string of the unit name for a given integer of INSUNITS, but in some environment it always returns `"unitless"` regardless of an input.

I guess this happens in __Python 3.11__ or later because since then the behaviour of `IntEnum.__str__` has been changed as follows:

https://docs.python.org/3/library/enum.html#enum.IntEnum
> Changed in version 3.11: [__str__()](https://docs.python.org/3/reference/datamodel.html#object.__str__) is now int.__str__() to better support the replacement of existing constants use-case. [__format__()](https://docs.python.org/3/reference/datamodel.html#object.__format__) was already int.__format__() for that same reason.

(I think [the implementation](
https://github.com/mozman/ezdxf/blob/c0a649d263724254b9c2176ee5c9280d81a1b163/src/ezdxf/units.py#L201-L206) of `unit_name` can just use `InsertUnits(enum).name` instead of `str()`.)

**To Reproduce**

```
print(ezdxf.units.unit_name(0))  # -> unitless
print(ezdxf.units.unit_name(1))  # -> unitless
```

As mentioned above, this should be reproduced in Python 3.11 (or later, but I haven't checked that).

**Expected behavior**

```
print(ezdxf.units.unit_name(0))  # -> Unitless
print(ezdxf.units.unit_name(1))  # -> Inches
```

privet-kitty:
fix #1131

### DIFF
--- a/src/ezdxf/units.py
+++ b/src/ezdxf/units.py
@@ -202,7 +202,7 @@ def unit_name(enum: int) -> str:
     """Returns the name of the unit enum."""
     try:
         return InsertUnits(enum).name
-    except (ValueError, IndexError):
+    except ValueError:
         return "unitless"
 
 
--- a/src/ezdxf/units.py
+++ b/src/ezdxf/units.py
@@ -201,7 +201,7 @@ def conversion_factor(
 def unit_name(enum: int) -> str:
     """Returns the name of the unit enum."""
     try:
-        return str(InsertUnits(enum)).split(".")[1]
+        return InsertUnits(enum).name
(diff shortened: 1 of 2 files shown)


mozman:
Thank you, the failing `mypy` test is not related to this issue.

PR #1155, opened 2024-08-30
Optional content groups in muPDF backend

mbway:
This PR introduces the option for the muPDF backend to tag entities as belonging to 'optional content groups' which are similar to layers in that visibility of entities in the groups can be toggled in supported viewers. [muPDF OCG documentation](https://pymupdf.readthedocs.io/en/latest/recipes-optional-content.html). This feature was requested in #1154 .

The current implementation only outputs visible layers as entities on invisible/hidden layers are filtered out by the frontend. A future improvement could be to use `--all-layers-visible` together with an option for the backend to set the layer visibility in the output independently from their visibility in the dxf.

Some fixes have also been applied to the `ezdxf draw` command as the `--all-layers-visible` had stopped working.

### DIFF
--- a/src/ezdxf/addons/drawing/pymupdf.py
+++ b/src/ezdxf/addons/drawing/pymupdf.py
@@ -3,16 +3,17 @@
 from __future__ import annotations
 
 import math
-from typing import Iterable, no_type_check
+from typing import Iterable, no_type_check, Any
 import copy
 
 import PIL.Image
 import numpy as np
 
-from ezdxf.math import Vec2, BoundingBox2d, Matrix44
+from ezdxf.math import Vec2, BoundingBox2d
 from ezdxf.colors import RGB
 from ezdxf.path import Command
 from ezdxf.version import __version__
+from ezdxf.lldxf.validator import make_table_key as layer_key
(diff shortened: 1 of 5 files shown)


src/ezdxf/addons/drawing/pymupdf.py
```diff
@@ -21,13 +22,13 @@
 from . import layout, recorder
 
 is_pymupdf_installed = True
+pymupdf: Any = None
 try:
-    import fitz
+    import pymupdf  # type: ignore[import-untyped]
```
mbway, line 27: fitz is a legacy name and mupdf is the new name: [source](https://pymupdf.readthedocs.io/en/latest/intro.html#note-on-the-legacy-module-name-fitz)

mbway:
the output of `ezdxf draw examples_dxf/visibility.dxf --backend mupdf --out out.pdf`

image

mozman:
I would not add support for exporting turned off or frozen layers, as this is often not the intention of the DXF file creator, although BricsCAD supports this as an option.

It is very laborious to implement and test and may break other functionality.

PR #1159, opened 2024-09-02
add native text rendering to muPDF backend

mbway:
This is an investigation into the feasibility of rendering text to PDF in a way that a viewer application is able to understand the text information (i.e. be able to select and copy the text). This was requested in https://github.com/mozman/ezdxf/discussions/1158 .

This PR introduces the `TextPolicy.NATIVE` setting to pass text information directly to the backend rather than 'baking' it to a series of glyph shapes since this looses the original text information.

If the backend supports text it can implement the `draw_text` method which gets called when `TextPolicy.NATIVE` is used. The PDF backend supports loading fonts and rendering text with arbitrary 2D transformations defined by a matrix.

The current implementation relies on a magic number to scale the font size correctly to match the baked text from the frontend (which I am treating as ground truth). For the files I have access to the value of 1.375 results in almost identical text however other fonts may require a different scale factor to get exact results.

In the following screenshot white is baked text and red is 'native' PDF text
image

### DIFF
--- a/src/ezdxf/addons/drawing/pipeline.py
+++ b/src/ezdxf/addons/drawing/pipeline.py
@@ -334,26 +334,33 @@ class RenderPipeline2d(AbstractPipeline):
             p.transform_inplace(transform)
         transformed_paths: list[BkPath2d] = glyph_paths
 
-        points: list[Vec2]
         if text_policy == TextPolicy.REPLACE_RECT:
-            points = []
-            for p in transformed_paths:
-                points.extend(p.extents())
-            if len(points) < 2:
+            bbox = self._text_bbox(transformed_paths)
+            if bbox is None:
                 return
-            rect = BkPath2d.from_vertices(BoundingBox2d(points).rect_vertices())
-            pipeline.draw_path(rect, properties)
+            pipeline.draw_path(BkPath2d.from_vertices(bbox.rect_vertices()), properties)
             return
         if text_policy == TextPolicy.REPLACE_FILL:
-            points = []
-            for p in transformed_paths:
-                points.extend(p.extents())
-            if len(points) < 2:
+            bbox = self._text_bbox(transformed_paths)
+            if bbox is None:
                 return
-            polygon = BkPoints2d(BoundingBox2d(points).rect_vertices())
             if properties.filling is None:
                 properties.filling = Filling()
-            pipeline.draw_filled_polygon(polygon, properties)
+            pipeline.draw_filled_polygon(BkPoints2d(bbox.rect_vertices()), properties)
+            return
+        if text_policy == TextPolicy.NATIVE:
+            bbox = self._text_bbox(transformed_paths)
+            if bbox is None:
+                return
+            abstract_font = self.text_engine.get_font(font_face)
+            self.backend.draw_text(
+                text,
+                bbox,
+                transform,
+                self.get_backend_properties(properties),
+                abstract_font,
+                cap_height,
+            )
             return
 
         if (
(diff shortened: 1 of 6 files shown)


mbway:
another possibility for the text rendering is to use `render_mode=3` which is invisible but allows the area where the text is to be selected. This is typically used for overlying OCR over scanned documents. This approach could be used if you prefer to always use the baked glyphs (because you can do advanced clipping etc) but still have the option to select the text. I think using `insert_text` instead of baking glyphs should result in a smaller file size though so may be desirable in some situations even if slightly less accurate.

mozman:
Sorry, but I will not add more complexity to the rendering process.

mbway:
Are you able to elaborate on what you don't like about this solution? Maybe I can find something that you like better?

src/ezdxf/addons/drawing/pymupdf.py
```diff
@@ -424,6 +429,46 @@ def draw_image(self, image_data: ImageData, properties: BackendProperties) -> No
             oc=self.get_optional_content_group(properties.layer),
         )
 
+    def register_font(self, font: AbstractFont) -> int:
```
mozman, line 432: I guess this implementation cannot handle SHX fonts.

mozman:
I liked the fact that the burden of rendering text in backends was removed. This feature is only optional, but questions will still arise as to why the text looks different with different backends.

This implementation skips the clipping stage and renders text outside of VIEWPORTs and clipped INSERT entities and of course cannot render SHX fonts.

I think this feature causes more problems than it solves.

src/ezdxf/addons/drawing/pymupdf.py
mbway: I would assume so. I wonder what autocad / other cad programs do when exporting if non-ttf fonts are used?

If both exact rendering and selectable/interpretable text was desirable then invisible text could be layered on top of the baked glyphs hypothetically

mbway:
I definitely see your viewpoint that users may be confused by the edge cases and that discarding text information before the backend results in simpler backends but I would think in some cases the ability to further process/analyze the output outweighs inaccuracies in rendering. I am not a user with this requirement though so I don't mind if we skip this feature.

It is a shame that without giving the backend more access, a user with this requirement cannot easily maintain a custom backend that handles text differently. I think a large restructuring would be required to allow this flexibility.

src/ezdxf/addons/drawing/pipeline.py
```diff
@@ -334,26 +334,33 @@ def draw_text(
...
                 properties.filling = Filling()
-            pipeline.draw_filled_polygon(polygon, properties)
+            pipeline.draw_filled_polygon(BkPoints2d(bbox.rect_vertices()), properties)
+            return
+        if text_policy == TextPolicy.NATIVE:
+            bbox = self._text_bbox(transformed_paths)
+            if bbox is None:
+                return
+            abstract_font = self.text_engine.get_font(font_face)
+            self.backend.draw_text(
```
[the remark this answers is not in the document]
mbway: Is clipping done in the pipeline? In which case I think you are correct, I didn't really handle clipping so I can't comment on if that would be difficult to add or not

mozman:
I created this tool with my needs in mind (I have been working with CAD for civil engineers for over 25 years) and I don't understand why users would want to create, edit and extract text in DXF files when that's what CAD applications are for. I have always been interested only in geometry stored in DXF files and automating geometry creation - an application independent scripting tool.

However, if someone wants to select/extract text from a DXF file I recommend a tool called `ezdxf`:smile::
```Python
import ezdxf

doc = ezdxf.readfile("your.dxf")
for entity in doc.query("MTEXT TEXT"):
    print(entity.dxf.text)  # or write it into a file
```

mbway:
There is some benefit in being able to access the text in its rendered form (i.e. position, colour etc) as DXF does not store this information plainly (hence the need for the rendering frontend). However, the use case for this may be niche and I'll let users who have an actual use case advocate for it if there are any.

Another benefit for letting PDF handle the text is smaller file sizes but again that may not be a priority.

Issue #1167, opened 2024-09-18
mline.py crashes trying to replace a Vec3 with ints instead of floats


Kaapp:
**Describe the bug**
When processing a particular file, I encountered a crash within `mline.py` where it tries to read the elevation using the `Vec3` replace function. It seems to be a very trivial fix, which I can raise a PR for if that's helpful.

**To Reproduce**
Information and data needed to reproduce the error:
1. A *simplified* and executable Python script which triggers the error.
Just trying to run the same code as the existing script will generate the error:
```
# original code in `render/mline.py#L41`
attribs["elevation"] = Vec3(ocs.from_wcs(bottom_border[0])).replace(
    x=0, y=0
)

# simplifies to:
border = Vec3(1.123, 2.123, 3.123)
border.replace(x=0, y=0)
```
3. Information about the *ezdxf* version and the OS.
Version 1.3.3 on Debian 12. (specifically we're using the `python:3.11-bookworm` docker image)
4. If processing a loaded DXF file causes the error, add a *simplified* and *zipped* 
   DXF file which still triggers the error.
I'd prefer to avoid sharing the file that caused this for confidentiality reasons, and because it can be replicate by calling the same Vec3 function with the provided parameters.

**Confidential Data**
If you have sensitive files that you can't disclose, try to remove everything 
sensitive from an existing file or create a brand new one which still triggers 
the error. Alternatively, you can send the zipped files to confidential@mozman.at, 
this data is only saved locally (no cloud upload of any kind) and the files are 
deleted after the error has been fixed.

**Expected behavior**
Code shouldn't crash for providing an int where a float was expected.

**Screenshots**

Kaapp:
thanks for merging the fix! 🙏

PR #1168, opened 2024-09-18
Fix mline.py crash due to passing int instead of float

Kaapp:
Based on [this bug report](https://github.com/mozman/ezdxf/issues/1167). I think this is a pretty trivial fix, but if I'm mistaken please do let me know!


mozman:
Thank you!

PR #1172, opened 2024-09-23
Overload `fast_plain_mtext` to different return types depending on `split=`

bswck-hai:
- [x] Type checked the code.

### DIFF
--- a/src/ezdxf/tools/text.py
+++ b/src/ezdxf/tools/text.py
@@ -9,11 +9,13 @@ from typing import (
     Iterable,
     Iterator,
     TYPE_CHECKING,
+    Literal,
     Union,
     Optional,
     Callable,
     NamedTuple,
     Any,
+    overload,
 )
 import enum
 import re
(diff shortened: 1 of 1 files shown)


bswck-hai:
@mozman Please review.

mozman:
Thanks for your contribution, but currently I integrate only bug fixes for more information see my latest posting #1175.

src/ezdxf/tools/text.py
```diff
@@ -9,11 +9,13 @@
     Iterable,
     Iterator,
     TYPE_CHECKING,
+    Literal,
     Union,
     Optional,
     Callable,
     NamedTuple,
     Any,
+    overload,
```
mozman, line 18: has to be imported from ` typing_extensions` , not available in Python 3.9 (requires 3.11)

src/ezdxf/tools/text.py
```diff
@@ -414,8 +416,18 @@ def plain_text(text: str) -> str:
...
+
+
+@overload
+def fast_plain_mtext(text: str, split: Literal[False] = False) -> str:
+    ...
+
+
 # pylint: disable-next=too-many-branches
-def fast_plain_mtext(text: str, split=False) -> Union[list[str], str]:
+def fast_plain_mtext(text: str, split: bool = False) -> Union[list[str], str]:
```
mozman, line 430: I don't add `bool` for such cases, type inference by `False` and `True`, shorter signatures are better, therefore return `list[str]|str` .

src/ezdxf/tools/text.py
bswck-hai: I believe you misread the documentation. `typing.overload` is the part of PEP 484.

src/ezdxf/tools/text.py
bswck-hai: I'd argue that shorter signatures aren't better when compared against actually accurate signatures.
Type inference from `False` is `Any` in mypy (not mentioning it erroring in strict mode).

src/ezdxf/tools/text.py
```diff
@@ -9,11 +9,13 @@
     Iterable,
     Iterator,
     TYPE_CHECKING,
+    Literal,
```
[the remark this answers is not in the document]
bswck-hai: Why? `Literal` has been there since Python 3.8, `ezdxf` supports 3.9+.

src/ezdxf/tools/text.py
bswck-hai: > therefore return `list[str]|str`
Migrating to PEP 604 is a topic for a different PR.

bswck-hai:
@mozman could you please reopen this?

PR #1174, opened 2024-09-29
Use `operator.itemgetter` instead of a custom lambda in `ezdxf.math.rtree.box_split`

bswck-hai:
Should generally be slightly faster. Lmk if you're interested in benchmarks backing up my thinking though.


mozman:
Thanks for your contribution, but currently I'm only integrating bug fixes. For more information, see my latest post #1175.

PR #1195, opened 2024-11-07
Insert.add_auto_attribs() support default values

gswifort:
It is probably intuitive to assign a default value from the block (attribute) definition.


mozman:
Thank you, again!

PR #1253, opened 2025-03-06
docs: Fix vport docs

lyzlisa:
This updates the docs to align with the attributes of the `VPort` entity used in code

https://github.com/mozman/ezdxf/blob/c1b01ca4b7baccb7751da23c4a1be2dfb5d9a35d/src/ezdxf/entities/vport.py#L44-L51


mozman:
Thank you!

PR #1267, opened 2025-03-25
Fix MultiLineString in GeoProxy

rdesparbes:
## Description

This pull request fixes a bug in `ezdxf.addons.geo` where the conversion to `MultiLineString` does not completely behave as expected: the coordinates in the resulting geometry are `Vec3` instances instead of simple `tuple`s, as indicated by the [`__geo_interface__`](https://gist.github.com/sgillies/2217756#__geo_interface__) specification.

This does not bother *shapely*, but some other libraries that support the `__geo_interface__` fail to parse the resulting geometry, like [geojson](https://github.com/jazzband/geojson) for example. 

## Environment used to reproduce

- OS: Windows 10
- Python 3.12
- _ezdxf_ 1.4.0

### DIFF
--- a/integration_tests/test_geo.py
+++ b/integration_tests/test_geo.py
@@ -61,5 +61,21 @@ def test_valid_hatch():
     assert p.root != {}
 
 
+def test_proxy_given_hatch_when_force_line_string_generates_valid_multi_line_string() -> None:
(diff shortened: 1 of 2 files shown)


mozman:
Thank you!

Issue #1286, opened 2025-04-22
cant read MULTILEADER as ProxyGraphic


zassoulin:
**Describe the bug**
in proxygraphic.py when trying to read MULTILEADER(atleast for me could also be due to corrupted buffer) struct.unpack_from(...) line 318 at unsafe_virtual_entities() fails to read the buffer and size, type_ are returned/set as 0 while raising ValueError thus causing an infinite loop.
the cause if the loop is size is set to 0 thus when iterating index +=size we get an infinite loop
the fix should be simple as handling the case where size is 0.
on a side note on note in line 316 buffer=self._buffer while the function unpack_from() recives self._buffer it dosent effect anything but seems like a minor human error which could be fixed along the way

**To Reproduce**
Information and data needed to reproduce the error:
1. unfortunately i am unable to supply the dxf file used and incase the information here wont be enough i will try and recreate it in a blank page, but as i spotted the issue i belive it should be enough.
2. the bug was found in windows with ezdxf versions 1.3.2 and 1.4.1

**Expected behavior**
Handel the MULTILEADER/corrupted buffer correctly and not be stuck on an infinite loop.
in my case the issue happened while using bbox.extents()

mozman:
Is this a valid proxy graphic/DXF file? Can you open the DXF file in AutoCAD or any other reliable CAD application?

zassoulin:
you can open it in abviewer so if it is corrupted abviewer can still open it but either way the fix is the same

PR #1287, opened 2025-04-22
suggested bug fix for 1286

zassoulin:
adding an edge case of jumping to end of buffer when buffer is failed to be read/unpacked.
i am unsure of where/if to add an unit test for the bug.
another possible solution is to instead of jump to end of buffer to jump a small increment such as 8.


mozman:
Thank you! 

I will add a check for chunk size < 1 in general, because this can not be correct and I assume the proxy graphic is corrupt or invalid.  #1286

PR #1292, opened 2025-05-02
Optimize shape handling in PyMuPdfRenderBackend

leonardopbatista:
Each drawing operation previously created and committed a new shape, causing significant performance overhead. This update introduces a shared shape instance reused across operations and committed once in finalize().

This change was inspired by a suggestion from the PyMuPDF maintainer in pymupdf/PyMuPDF#3800. Performance improvements can exceed 100x in documents with many entities.


mozman:
Thank you!

The file size of PDFs appears to have increased, but that is a small price to pay for the enormous performance gain.

mozman:
I was wrong, the size of the new PDFs is also smaller! In average: 38x faster and 2.5x smaller!

image

Issue #1313, opened 2025-07-06
DEVISION BY 0 in draw_image_entity


zassoulin:
**Describe the bug**
when drawing a layout (FrontEnd.(draw_layout))
at ezdxf\addons\drawing\frontend.py line 775 an devision by 0 occures.
it happens when text is empy/white space (line 769 text = image_def.dxf.filename)
front.text_wudth_ex(...) has a check that returns text_with = 0 incase of white space thus causing a devesion by 0

**Expected behavior**
Still RUN even is such as edge case.
a couple options are availabe for such an edge case such as inserting an '<UNKOWN IMAGE NAME>' . 
or not drawing if it is a white space i will provide a merge request to help solve the issue

PR #1314, opened 2025-07-06
fixing devision by 0 and improving cache call

zassoulin:
fixxed issue where incase of whitespace file name devision by 0 would occuer see issue #1313.
now when a file name is whitespace no text would be rendered.
also it may be better to use .get() instead of try/catch in ttfonts as try/catch are usualy not ideal.


src/ezdxf/fonts/ttfonts.py
mozman, line 1: This is just an opinion and I will ignore this.

src/ezdxf/addons/drawing/frontend.py
mozman, line 1: This is a valid fix and I will implement this.

PR #1317, opened 2025-07-31
fix(#1316): Filter polyline vertices for rendering based on vertex flags

### LINKED ISSUE #1316: 2D-Polyline (AcDb2dPolyline) gets rendered to svg incorrectly
**Describe the bug**

When rendering a dxf file to svg, AcDb2dPolyline elements (in this case with cubic) get rendered incorrectly.

Looking at the generated path it seems like there are additional points added to those needed for the outline of the line.

**To Reproduce**
Information and data needed to reproduce the error:

Python              - 3.12.5
ezdxf                 - 1.4.2
MacOs Silicon

```
import ezdxf
from ezdxf.addons.drawing import RenderContext, Frontend, layout, svg

class DXFToSVGConverter:
    def __init__(self, dxf_file_path):
        self.doc = ezdxf.readfile(dxf_file_path)
        self.msp = self.doc.modelspace()

    def convert_to_svg(self, output_path):
        backend = svg.SVGBackend()
        Frontend(RenderContext(self.doc), backend).draw_layout(self.msp)

        with open(output_path, "wt") as fp:
            fp.write(backend.get_string(layout.Page(0, 0)))


def convert_dxf_to_svg(dxf_path, svg_path):
    converter = DXFToSVGConverter(dxf_path)
    converter.convert_to_svg(svg_path)

if __name__ == "__main__":
    dxf_file = "Drawing.dxf"
    
    convert_dxf_to_svg(dxf_file, "Drawing.svg")
```

**Expected behavior**
The line should be displayed as one curved line and not look like it does

**DXF File to reproduce as well as the resulting svg/png**
[Archive.zip](https://github.com/user-attachments/files/21515141/Archive.zip)

PaulLampe:
fixes #1316
Add vertex filtering to exclude auto-generated curve-fit vertices and spline frame control points from rendering polylines as they should not be included visually.

### DIFF
--- a/src/ezdxf/render/trace.py
+++ b/src/ezdxf/render/trace.py
@@ -411,6 +411,37 @@ class CurvedTrace(AbstractTrace):
             vtx1 = vtx2
 
 
+def should_include_vertex_for_rendering(vertex_flags: int) -> bool:
+    """Determine if a vertex should be included in rendering based on its flags.
+
+    According to the DXF specification:
+    - Flag 0: Straight-segment vertex (include)
+    - Flag 1: Extra curve-fit vertex (exclude - auto-generated)
+    - Flag 2: Curve-fit tangent vertex (include - carries tangent info)
+    - Flag 8: Spline-vertex (fit-point) for 2D spline (include)
+    - Flag 16: Spline-frame control point (exclude - internal calculation only)
+
+    Args:
+        vertex_flags: The vertex flags value (group code 70)
+
+    Returns:
+        True if the vertex should be included in rendering, False otherwise
+    """
+    # Exclude extra vertices created by curve-fitting (flag 1)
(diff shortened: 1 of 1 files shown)


src/ezdxf/render/trace.py
```diff
@@ -606,3 +613,33 @@ def from_polyline(cls, polyline: DXFGraphic, segments: int = 64) -> TraceBuilder
             # miter at the closing point. (only linear to linear trace).
             trace.close()
         return trace
+
+    def should_include_vertex_for_rendering(vertex_flags: int) -> bool:
```
mozman: @staticmethod is missing or implement as a simple function, doesn't have to be a class-method.

PaulLampe: Moved it to before the class, does that work for you? Sorry, not at all a python guy 😅

Issue #1318, opened 2025-08-01
Invalid viewport offset relative layout


eXponenta:
**Describe the bug**
Viewport content bring shift when viewports more than 1 on same layout.

I use "ezdxf view .." api for validate ezdxf parsing stability.

**To Reproduce**
Looks Layer3 layout in attachment.

I drawn blue rectangle around required extents of viewports content.

[Gasket.dxf.zip](https://github.com/user-attachments/files/21548513/Gasket.dxf.zip)

<img width="862" height="618" alt="Image" src="https://github.com/user-attachments/assets/1d9c1807-b195-4c77-b5cb-38a1b01ef879" />

**Expected behavior**

Viewport content not shifts.

<img width="1135" height="798" alt="Image" src="https://github.com/user-attachments/assets/002adfb8-e7ab-4ee4-8b38-6b03cbad2dca" />

eXponenta:
This happens because there are ambiguity with **center_point** and **target_point** of viewport.

valid pair of CP and TP:
<img width="916" height="784" alt="Image" src="https://github.com/user-attachments/assets/f1c83289-8247-44ae-a0c4-dcdc60be6951" />

Invalid pair:

<img width="728" height="769" alt="Image" src="https://github.com/user-attachments/assets/f22333d8-d502-4d2e-bb2d-c06d0979333c" />

( shifts presented only in ezdxf, TrueDWG and other render blue circle  exactly in cross of lines ) 

This happens when you edit model view inside viewport ( any changes shift CP and TP of viewport ) but not change viewport clip ( you can shift viewport sizes little bit and restore then back and ezdx will got valid CP).

https://github.com/mozman/ezdxf/blob/e5b98c73e8f16ea6f03e7fe35d141b1d64d40602/src/ezdxf/entities/viewport.py#L636

PR #1319, opened 2025-08-01
fix: take accoun target point and always return center_point

eXponenta:
Try fix invalid viewport transforms.
#1318


eXponenta:
Made update for test, because older has invalid extents in modelspace after calculation.
see attached DXF.

[twisted_viewport_modelspace_limits.dxf.zip](https://github.com/user-attachments/files/21561011/twisted_viewport_modelspace_limits.dxf.zip)

mozman:
Thank you for you contribution. 

This may take a while until I will merge this ... this is a topic without much documentation and requires therefore laborious manual testing with examples files I have collected over time.

PR #1320, opened 2025-08-01
fix: render viewport as regular entity

eXponenta:
Viewport is graphics like a polyline with stroke.
Also viewports should render in valid order for allow place entities over, like annotation, hatch, texts etc.

Current implementation:
Снимок экрана 2025-08-01 в 22 01 55

PR:
Снимок экрана 2025-08-01 в 22 08 59


eXponenta:
Test (render_over_viewport layout)
[Gasket.dxf.zip](https://github.com/user-attachments/files/21553491/Gasket.dxf.zip)

PR #1343, opened 2025-11-06
Add Cython optimization for ASCII DXF tag loading

adacovsk:
## Summary

Implements Cython-optimized versions of `ascii_tags_loader()` and `tag_compiler()` for improved performance when reading large ASCII DXF files.

## Use Case

I'm working with 6+ GB ASCII DXF files that currently require approximately 16 minutes to load.

## Benchmark Results

**150MB Polyface Mesh File:**
- Pure Python: 60.35 seconds
- Cython: 31.71 seconds  
- Speedup: 1.90x

**164MB Line Entities File:**
- Pure Python: 37.31 seconds
- Cython: 20.07 seconds
- Speedup: 1.86x

## Implementation Details

- Uses C-level string parsing (`atoi`, `strtod`) instead of Python functions
- Inline type checks instead of dictionary lookups
- Automatic fallback to pure Python when Cython unavailable
- Fully backward compatible

## Changes

- `src/ezdxf/acc/tagger.pyx`: New Cython implementation
  - `ascii_tags_loader()`: Fast tag reader using C string functions
  - `tag_compiler()`: Fast type compilation with inline checks
- `src/ezdxf/lldxf/tagger.py`: Modified to dispatch to Cython when available
- `setup.py`: Added tagger extension to build list
- `tests/test_09_cython_acceleration/test_907_acc_tagger.py`: 14 new tests

## Testing

- ✅ All 22 existing tests pass (test_042_tags_loader.py)
- ✅ All 14 new Cython-specific tests pass
- ✅ Verified: Cython produces identical output to Python

## Usage

The optimization is automatically used when Cython is available:

```python
import ezdxf

# Automatically uses Cython if compiled
doc = ezdxf.readfile("large_file.dxf")
```

To build:
```bash
pip install Cython
python setup.py build_ext --inplace
```


mozman:
I had the same idea a long time ago, but I figured if most of the work involved file I/O, Python might not be the bottleneck, so I didn't pursue it further.

If your work confirms this assumption, you may have saved me a lot of work. Thanks.

PR #1350, opened 2025-11-26
fix: valid float parsing when pattern not include leading number before . ( .231 )

eXponenta:
in some DWGs (and then DXFs after ODA transform ) MTEXT include width pattern as \W.85; instead of \W0.85;

Снимок экрана 2025-11-26 в 14 36 44
(See the attached file.)

And this text will present then in text content.

Before:
Снимок экрана 2025-11-26 в 14 45 46

After:
Снимок экрана 2025-11-26 в 14 47 56

file:
[mtext_invalid_char_width.dxf 2.zip](https://github.com/user-attachments/files/23769346/mtext_invalid_char_width.dxf.2.zip)

### DIFF
--- a/tests/test_05_tools/test_521_mtext_parser.py
+++ b/tests/test_05_tools/test_521_mtext_parser.py
@@ -394,6 +394,9 @@ class TestMTextContextParsing:
             r"\H0.3e1;",
             r"\H30e-1",
             r"\H30e-1;",
+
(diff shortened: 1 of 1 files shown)


eXponenta:
Now we got problem with test, because it think that `.{d}` is invalid instruction, when for cad it is ok

Issue #1351, opened 2025-11-28
Inconsistent text layout when text style non-default


eXponenta:
**Describe the bug**
If we have a text style that already expose `width_factor`, `line_space_factor` etc then when you use MTEXT  it won't applied at all.

This happens because ezdxf not pass text style to backend at all and fonts emit shapes without apply a text style.

MTEXT in composite mode use a layout engine, that resolve style from instruction set, BUT also this isn't fully correctly and not all CAD emit default \W and H instruction when it same as text style.

**Expected behavior**
Stable text layout because it corrupt some text when it placed by word (see top image ).

**Screenshots**
From `ezdx view` on latest version:

Samples:

<img width="500" height="360" alt="Image" src="https://github.com/user-attachments/assets/aab026f3-8440-4227-805f-b8e14dbaa842" />

From DWG TrueView 2020 ( expected, same and in NANOCAD ):

<img width="500" height="240" alt="Image" src="https://github.com/user-attachments/assets/36d8a657-c2df-46bd-adc1-768e66f94af0" />

**Source**:

[inconsist_text_emit_2.dxf.zip](https://github.com/user-attachments/files/23827642/inconsist_text_emit_2.dxf.zip)

PR #1353, opened 2025-11-28
feat: valid resolve style width factor in backend

eXponenta:
Fix: #1351 

Resolve TextStyleProperties which pased to backend and then used for compute layout.

#1350 is required for stable parsing irregular instructions.

### DIFF
--- a/src/ezdxf/addons/drawing/text.py
+++ b/src/ezdxf/addons/drawing/text.py
@@ -321,12 +321,12 @@ def simplified_text_chunks(
 
     cap_height = _get_cap_height(text)
 
-    extra_scale = 1
+    extra_scale = 1.0
     # get layout with from text when it simple text or from MTEXT style 
     if text_style and isinstance(text, MText):
         extra_scale = text_style.width
     elif isinstance(text, Text):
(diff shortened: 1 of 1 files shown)


eXponenta:
Looks like normal

Снимок экрана 2025-11-28 в 20 51 22

eXponenta:
@mozman 

So, this means that any MTEXT is not simply? Because any MTEXT has style reference, which can change MTEXT transformation rules, otherwise need check that width factor is 1, oblique is 0 of refered style.

By doc  a Simple MText is text without inline layout codes ( w, , q, p etc ) that requires parsing and evaluation per instruction set ( which also can be exploded to simple MText ).

Yep, if you want I can add style validation and forward it  to complex MText renderer, but table style also need to pass anyway.

mozman:
The `MTextContext` has a `width_factor` attribute:

https://github.com/mozman/ezdxf/blob/d9263cf3a427b7d09d3bdabaf4674cb1c4059544/src/ezdxf/tools/text.py#L1016

Text styles and MText styles can be accessed through the `MText.doc` attribute if not `None`.

PR #1374, opened 2026-02-18
Fix Body.copy_data() sharing _temporary_transformation between original and copy

### LINKED ISSUE #1373: Body.copy_data() shallow-copies _temporary_transformation, corrupting originals on copy+transform
## Bug

`Body.copy_data()` shares the same `TransformByBlockReference` object between original and copy. When the copy is transformed (e.g., via `bbox.extents()` → `virtual_entities()` → `copy()` → `transform()`), the original entity gets a spurious pending transform. On save, `transform_entity_by_blockref()` moves each contaminated entity into a new anonymous `*U` block, destroying block structure.

**One-line root cause** in `ezdxf/entities/acis.py`:
```python
entity._temporary_transformation = self._temporary_transformation  # shared reference
```

## Reproduction

Any DXF with SURFACE/REGION entities inside a block referenced via INSERT:

```python
import ezdxf
from ezdxf import bbox

doc = ezdxf.readfile("file_with_acis_in_blocks.dxf")

# Verify clean state
for block in doc.blocks:
    for e in block:
        if hasattr(e, '_temporary_transformation'):
            assert e._temporary_transformation.get_matrix() is None  # passes

bbox.extents(doc.modelspace())

# Originals are now contaminated
for block in doc.blocks:
    for e in block:
        if hasattr(e, '_temporary_transformation'):
            if e._temporary_transformation.get_matrix() is not None:
                print(f"BUG: {e.dxftype()} in '{block.name}' has pending transform")
```

After `saveas()` + reload, SURFACE/REGION entities are replaced by INSERT references to anonymous `*U` blocks.

Tested with a manufacturer-provided 3D connector block (Balluff BNI005M) containing 19 SURFACE + 18 REGION entities — all 37 were decomposed after `bbox.extents()` + save.

## Proposed Fix

Replace the shallow assignment with a new instance:

```python
entity._temporary_transformation = TransformByBlockReference()
```

I have a fix with tests ready and will open a PR shortly.

BreenMac:
Fixes #1373

`Body.copy_data()` assigned the same `TransformByBlockReference` to the copy. Transforming the copy contaminated the original, destroying block structure on save.

**Fix:** Create a new `TransformByBlockReference` for the copy, preserving any pending matrix value.

```python
# Before (shared reference)
entity._temporary_transformation = self._temporary_transformation

# After (independent copy)
entity._temporary_transformation = TransformByBlockReference()
m = self._temporary_transformation.get_matrix()
if m is not None:
    entity._temporary_transformation.set_matrix(m.copy())
```

**Test added:** `test_copy_has_independent_temporary_transformation` — verifies the copy and original have independent transformation objects and that transforming the copy leaves the original unaffected.

All existing tests pass (8220 passed, 52 skipped).

### DIFF
--- a/src/ezdxf/entities/acis.py
+++ b/src/ezdxf/entities/acis.py
@@ -144,7 +144,12 @@ class Body(DXFGraphic):
         entity.sat = self.sat
         entity.sab = self.sab  # load SAB on demand
         entity.dxf.uid = guid()
-        entity._temporary_transformation = self._temporary_transformation
+        # Create a new TransformByBlockReference for the copy — sharing the same
+        # object causes transforming the copy to contaminate the original.
+        entity._temporary_transformation = TransformByBlockReference()
+        m = self._temporary_transformation.get_matrix()
+        if m is not None:
+            entity._temporary_transformation.set_matrix(m.copy())
 
     @override
     def map_resources(self, clone: Self, mapping: xref.ResourceMapper) -> None:
--- a/tests/test_02_dxf_graphics/test_232e_acis_transform.py
+++ b/tests/test_02_dxf_graphics/test_232e_acis_transform.py
@@ -71,5 +71,25 @@ def test_transform_and_copy(body: Body):
     assert m.transform(v).isclose(m2.transform(v))
 
 
+def test_copy_has_independent_temporary_transformation(msp: Modelspace):
+    """Copy and original must not share the same TransformByBlockReference.
+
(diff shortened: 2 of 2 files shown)


mozman:
Thank you! 

The diagnosis is correct, but the fix is not what I want. 
The temp transformation  should only be applied when the DXF is exported.
The `temporary_transformation` must be copied.

PR #1381, opened 2026-05-04
feat: add content_hash property to DXFEntity

shplishka:
## Summary

- Adds a `content_hash` property to `DXFEntity` that returns a deterministic SHA-256 hex digest derived from the entity's DXF content (entity-specific attributes + xdata).
- The hash is **identity-agnostic**: handle, owner, and Python object identity are excluded, so two entities with identical DXF properties produce the same hash.
- Useful for **duplicate detection**, **content change tracking**, and **diffing** DXF documents.

## Implementation

The property serializes entity content via the existing `export_entity()` and `export_xdata()` methods into a `TagWriter` with `write_handles=False`, then hashes the resulting string with SHA-256. This leverages the same serialization path used for DXF file export, so it captures all entity data including complex sub-structures (LWPOLYLINE vertices, HATCH paths, SPLINE control points, MTEXT text content, etc.).

The hash is computed on demand and **not cached**, because any attribute mutation would silently invalidate a cached value.

### Example usage

```python
import ezdxf

doc = ezdxf.readfile("drawing.dxf")
msp = doc.modelspace()

seen = set()
for entity in msp:
    h = entity.content_hash
    if h in seen:
        print(f"Duplicate: {entity}")
    seen.add(h)
```

## Test plan

- [x] `content_hash` returns a 64-character hex string (SHA-256)
- [x] Entities with identical properties produce the same hash
- [x] Entities with different properties produce different hashes
- [x] Hash ignores handle differences
- [x] Hash ignores owner differences
- [x] Hash changes when entity attributes are mutated
- [x] Different layers produce different hashes
- [x] A copy produces the same hash as the original
- [x] Hash is deterministic (same result on repeated calls)
- [x] Works on base `DXFEntity` class (not just subclasses)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/ezdxf/entities/dxfentity.py
+++ b/src/ezdxf/entities/dxfentity.py
@@ -27,6 +27,7 @@ from typing import (
 )
 from typing_extensions import Self
 
+import hashlib
 import logging
 import uuid
 from ezdxf import options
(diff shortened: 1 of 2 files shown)


mozman:
Thanks for your contribution.

Here some quick thoughts before I decide if I will merge this feature:

1. Do not put computationally expensive features like this into `properties`, `properties` should be fast and simple.
2. This feature does not need to be part of the class. A simple function does the job as well.
3. Some entities also use extension dictionaries to store data like `MTEXT` columns.
4. `INSERT` refer block definition by name; block definitions with the same name in different documents may not contain the same content.

I am really not sure if this is useful in practice.

PR #1383, opened 2026-05-09
Refactor clipping logic in UniversalFrontend and add stack depth prop…

origami7:
…erty to ClippingPortal

- Improved the conditional structure for clipping in UniversalFrontend to enhance readability.
- Introduced a new property `stack_depth` in ClippingPortal to track the number of clipping stages.
- Added tests to ensure push/pop balance for clipping operations in various scenarios.


mozman:
Thank you, the mypy errors are caused by a new defaults in mypy and are already fixed by myself.

PR #1394, opened 2026-06-17
Enhance pattern transformation in DXFPolygon and OCSTransform

origami7:
- Implemented 2D direction transformation in OCSTransform for pattern offsets and line directions.
- Updated DXFPolygon to correctly scale and rotate pattern attributes without translation.
- Added comprehensive tests to verify the behavior of pattern transformations, including scaling, rotation, and the preservation of geometry during transformations.

### DIFF
--- a/src/ezdxf/entities/polygon.py
+++ b/src/ezdxf/entities/polygon.py
@@ -400,13 +400,14 @@ class DXFPolygon(DXFGraphic):
         )
         dxf.extrusion = ocs.new_extrusion
         if self.pattern:
-            # todo: non-uniform scaling
-            # take the scaling factor of the x-axis
-            factor = ocs.transform_length((self.dxf.pattern_scale, 0, 0))
             angle = ocs.transform_deg_angle(self.dxf.pattern_angle)
-            # todo: non-uniform pattern scaling is not supported
-            self.pattern.scale(factor, angle)
-            self.dxf.pattern_scale = factor
+            orig_scale = self.dxf.pattern_scale
+            new_scale_factor = ocs.transform_length((1.0, 0, 0))
+            for line in self.pattern.lines:
+                line.base_point = ocs.transform_2d_direction(line.base_point)
+                line.offset = ocs.transform_2d_direction(line.offset)
+                line.angle = (line.angle + angle) % 360.0
+            self.dxf.pattern_scale = orig_scale * new_scale_factor
             self.dxf.pattern_angle = angle
         self.post_transform(m)
         return self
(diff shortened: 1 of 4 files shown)


origami7:
### Description

**Describe the bug**
When a `HATCH` entity is nested inside an `INSERT` (Block Reference) that has a rotation angle (e.g., 180 degrees), the `Frontend` rendering pipeline fails to render any hatch lines (outputting an empty/transparent filling). 

Through deep debugging, we found that the root cause lies in `DXFPolygon.transform()` (located in `src/ezdxf/entities/polygon.py`). While it correctly transforms the boundary paths, it **fails to properly apply the rotation and translation matrices to the hatch pattern's `base_point`**, applying only the pattern scale factor instead. This causes the pattern origin to fly millions of units away from the actual boundary, causing the internal ray-casting algorithm (`hatch_lines()`) to yield 0 line segments due to massive signed distance divergence.

### Mathematical Proof (Debug Evidence)

Here is the exact state comparison captured from a real failing case where the `INSERT` has a **180° rotation** and a translation matrix $T = (962.99, 969.88)$:

#### 1. Boundary Paths (Correctly Transformed)
* **Original vertex in block definition**: `(-64.92, 173.74)`
* **Transformed vertex in WCS**: `(1027.91, 796.14)` 
* *Calculation*: $180^\circ \text{ rotation} + T \implies (962.99 - (-64.92), 969.88 - 173.74) = (1027.91, 796.14)$  **(✓ Correct)**

#### 2. Pattern Base Point (Buggy Transformation)
* **Original pattern base_point**: `(-38543.80, -6670.83)`
* **Expected WCS base_point (after 180° rotation + Translation + Scale 200)**: 
  $$\begin{aligned} X &= (-(-38543.80) + 962.99) \times 200 = 7,901,358 \\ Y &= (-(-6670.83) + 969.88) \times 200 = 1,528,142 \end{aligned}$$
* **Actual Buggy Result produced by `transform()`**: `(7708760.44, 1334166.08)`
* *Dissection*: The code mistakenly skipped the rotation matrix and translation, evaluating to roughly $\text{base\_point} \times \text{scale}$, shifting the emission origin away by a delta of millions of units.

#### 3. Consequence on `hatch_lines()`
* **Signed distance from boundary to baseline**: `-4,507,355`
* **Pattern line spacing (`normal_distance`)**: `5,000`
* Since the gap is insanely huge, the ray-tracer skips the entire bounding box area: `Total hatch lines from virtual: 0`.

### Root Cause Location

The bug is situated inside:
📂 **`src/ezdxf/entities/polygon.py`** in method `DXFPolygon.transform(self, m: Matrix44)`.

**What the code currently does:**
It scales the pattern properties but mistakenly treats the `base_point` vector as a pure direction or invariant point, omitting the full affine transformation matrix $M$ (Rotation + Translation) of the parent `INSERT`.

**What the code should do:**
The `base_point` of a `HatchPatternLine` must undergo the exact same spatial transformation chain as the boundary vertices:
```python
# Conceptual fix inside DXFPolygon.transform()
for line in self.pattern.lines:
    # Ensure rotation and translation from Matrix44 are fully applied to the point
    line.base_point = m.transform(line.base_point)

PR #1398, opened 2026-07-05
fix: numerically robust ConstructionRay.intersect fixes aligned-dimension rendering for near-vertical segments

### LINKED ISSUE #1300: Incorrect aligned dimension extension lines and measurment text
For certain segments, rendering an aligned dimension results in incorrect extension lines and wrong measurement text. Depending on the point coordinates, it might be correct or wrong, while the `dimension.get_measurement()` returns the correct value.

Here is the reproducer example:
```python
import ezdxf
print(f"Using ezdxf version: {ezdxf.__version__}")

points = [
    (250235.3373553778, 99368),
    (250235.3373553779, 99412),
]

output_filename = "polyline_example.dxf"

doc = ezdxf.new()
doc.modelspace().add_lwpolyline(points)

p1, p2 = points[0], points[1]
doc.modelspace().add_aligned_dim(p1, p2, distance=20).render()

print(f"Polyline points: {points}")
dim = list(doc.modelspace().query("DIMENSION"))[0]
print(f"Dimension text: {dim.dxf.text}")
print(f"Dimension measurement: {dim.get_measurement()}")
for e in dim.virtual_entities():
    if e.dxftype() == "MTEXT":
        print(f"MTEXT text: {e.dxf.text}")

doc.saveas(output_filename)
print(f"DXF file '{output_filename}' created successfully!")
```
Note that the distance between the points is ~44 units.

I use python version 3.12.9

Output:
```
Using ezdxf version: 1.4.2
Polyline points: [(250235.3373553778, 99368), (250235.3373553779, 99412)]
Dimension text: <>
Dimension measurement: 44.0
MTEXT text: 32
DXF file 'polyline_example.dxf' created successfully!
```

Image

If I change the numeric values ever so slightly, I get the expected parallel lines and correct measurement text:
```
Using ezdxf version: 1.4.2
Polyline points: [(250235.337355377, 99368), (250235.337355377, 99412)]
Dimension text: <>
Dimension measurement: 44.0
MTEXT text: 44
DXF file 'polyline_example.dxf' created successfully!
```

Image

mvanhorn:
## Summary

Aligned dimensions whose two points are only slightly non-vertical (a tiny x-difference) now render the correct measurement and extension-line geometry instead of losing precision. `ConstructionRay.intersect` now computes the general-case intersection with a parametric determinant relative to the ray location rather than the slope / y-intercept form.

## Why this matters

Reported in #1300: `add_aligned_dim((250235.3373553778, 99368), (250235.3373553779, 99412), distance=20).render()` draws a measurement of `32` even though `dimension.get_measurement()` correctly returns `44`.

The dimension-line ray for an aligned dimension is near vertical, so `direction.x` is about `2.6e-12`, just above `ABS_TOL`, and the ray is treated as non-vertical with a slope near `3.78e11`. The general branch of `ConstructionRay.intersect` then solved `x = (ray1._yof0 - ray2._yof0) / (ray2._slope - ray1._slope)`, where `_yof0` is the y value at `x=0`, far from the actual coordinates (`x` around `250235`) and amplified by the huge slope. The result is catastrophic cancellation: the extension-ray intersections land at `99376` and `99408` (span `32`) instead of `99368` and `99412` (span `44`), which is exactly the value the renderer draws.

The fix replaces that computation with a parametric intersection expressed relative to `ray1._location`: with `o = ray2._location - ray1._location`, `denom = d1.x*d2.y - d1.y*d2.x`, the hit point is `ray1._location + d1 * ((o.x*d2.y - o.y*d2.x)/denom)`. Because everything is computed relative to a nearby location instead of `x=0`, there is no slope amplification and the result stays accurate for large coordinates and near-axis-aligned rays. The exact vertical/horizontal short-circuit branches and the `is_parallel` guard are unchanged (the guard ensures `denom` is non-zero here), and `_slope`, `_yof0`, `yof`, and `xof` are untouched, so this is a strict accuracy improvement with no API change.

## Testing

- Added a regression test in `test_641_construction_ray.py` that intersects a near-vertical ray at large coordinates with its orthogonal rays and asserts the span is `44`; it fails on the old slope/intercept code and passes with the fix.
- Added a symptom-level regression test in `test_704_render_linear_dimension.py` that builds the reporter's aligned dimension and asserts the rendered MTEXT measurement equals `44` and matches `get_measurement()`.
- `pytest tests/test_06_math tests/test_07_render` passes (2795 passed, 2 skipped, 1 xfailed); `mypy --ignore-missing-imports` on the changed module reports no issues.

Fixes #1300

### DIFF
--- a/src/ezdxf/math/line.py
+++ b/src/ezdxf/math/line.py
@@ -151,11 +151,17 @@ class ConstructionRay:
             y = ray2._location.y
             x = ray1.xof(y)
         else:
-            # calc intersection with the 'straight-line-equation'
-            # based on y(x) = y0 + x*slope
-            # guards above guarantee that no slope is None
-            x = (ray1._yof0 - ray2._yof0) / (ray2._slope - ray1._slope)  # type: ignore
-            y = ray1.yof(x)
+            # Parametric intersection relative to ray1._location. Computing the
+            # hit point relative to a nearby location (instead of the slope /
+            # y-intercept form based on x=0) avoids catastrophic cancellation
+            # for near-axis-aligned rays with large coordinates.
+            # The is_parallel() guard above guarantees denom != 0 here.
+            o = ray2._location - ray1._location
+            d1 = ray1._direction
+            d2 = ray2._direction
+            denom = d1.x * d2.y - d1.y * d2.x
+            t = (o.x * d2.y - o.y * d2.x) / denom
+            x, y = ray1._location + d1 * t
         return Vec2((x, y))
 
     def orthogonal(self, location: UVec) -> ConstructionRay:
--- a/tests/test_06_math/test_641_construction_ray.py
+++ b/tests/test_06_math/test_641_construction_ray.py
@@ -121,6 +121,26 @@ class TestConstructionRay:
         with pytest.raises(ParallelRaysError):
             _ = ray1.bisectrix(ray3)
 
+    def test_near_vertical_ray_large_coordinates(self):
+        # Regression test for issue #1300: a near-vertical ray with large
+        # coordinates (as produced by aligned dimensions) must intersect its
+        # orthogonal rays accurately. The old slope / y-intercept formula in
+        # intersect() suffered catastrophic cancellation here and returned a
+        # wrong span (32 instead of 44).
+        p0 = (250235.337, 99368)
+        p1 = (250235.337, 99412)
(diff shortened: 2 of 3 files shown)


mvanhorn:
Thanks for the merge, @mozman - robust ray intersection beats the epsilon dance.

mvanhorn:
Appreciate the review and merge, @mozman - glad the numerically robust ConstructionRay.intersect landed.

Issue #227, opened 2020-09-08
Incorrect layer color 7 resolving in drawing add-on


mozman:
DXF entities using ACI color `BYLAYER` and a layer with ACI color 7 (white/black) should get a dark color if the background is light and vice versa.

The current implementation does not work correct, because the layer color is fixed in the constructor of the `RenderContext()` object, the current layout is always the modelspace with a dark background color.

The new implementation should also support manual background color overriding by the package-user: `RenderContext.current_layout.set_colors()`

mozman:
New implementation works:

Black background:
![fillings_bg_black](https://user-images.githubusercontent.com/4230966/92554621-0ef8e500-f266-11ea-8608-f4c13ad821e4.png)

White background:
![fillings_bg_white](https://user-images.githubusercontent.com/4230966/92554635-17512000-f266-11ea-8eed-6cc7490b46b2.png)

Black transparent background

@mbway Is it possible in matplotlib to save PNG images with transparent background?
This image was created with background color "#00000000", but transparency is not present in image editor, no checker board as background visible:

![fillings_bg_transparent](https://user-images.githubusercontent.com/4230966/92555098-369c7d00-f267-11ea-9fe6-845a0d4f759a.png)

mbway:
yes, it is possible to have transparent and translucent backgrounds in matplotlib:
```python
import matplotlib.pyplot as plt

fig = plt.figure()
ax = fig.add_axes((0, 0, 1, 1))


# already handled inside matplotlib backend constructor:
ax.xaxis.set_visible(False)
ax.yaxis.set_visible(False)
for s in ax.spines.values():
    s.set_visible(False)
##################


ax.plot([0, 1], [0, 1])
ax.set_facecolor('#ff000055')

fig.savefig('test.png', facecolor=ax.get_facecolor(), transparent=True)
```
results in this image (which is semi-transparent):
![test](https://user-images.githubusercontent.com/4923501/92578516-d962f480-f283-11ea-90db-6da0099bd062.png)

mozman:
Perfect, thank you!
![fillings_bg_transparent](https://user-images.githubusercontent.com/4230966/92579147-1e3f5900-f28d-11ea-96c2-a5fff621bcba.png)

PR #1401, opened 2026-08-01
Honour plot style table color for ACI 7

rafaelfrancisco-contato:
# Honour plot style table color for ACI 7

## Problem

`RenderContext(doc, ctb=...)` applies the plot style table color for every ACI
**except 7**. Since ACI 7 is the default color of layer `0` and by far the most common
color in engineering drawings, a CTB that maps ACI 7 to black is silently ignored and
those entities render white — invisible when plotting onto white paper.

Reproducer with the current code:

```python
import tempfile
from pathlib import Path

import ezdxf
from ezdxf.addons import acadctb
from ezdxf.addons.drawing import RenderContext

tmp = Path(tempfile.mkdtemp())
ctb = tmp / "all_red.ctb"
table = acadctb.new_ctb()
for aci in range(1, 256):
    table[aci].color = (255, 0, 0)     # every ACI plots red
table.save(str(ctb))

doc = ezdxf.new("R2018", setup=True)
msp = doc.modelspace()
for aci in (1, 2, 7, 8):
    msp.add_line((0, aci), (10, aci), dxfattribs={"color": aci})

ctx = RenderContext(doc, ctb=str(ctb))
for entity in msp:
    print(entity.dxf.color, ctx.resolve_color(entity)[:7])
```

```
1 #ff0000
2 #ff0000
7 #ffffff   <-- expected #ff0000
8 #ff0000
```

The practical case is BYLAYER on layer `0`, which defaults to ACI 7:

```python
msp.add_line((0, 0), (10, 0), dxfattribs={"layer": "0", "color": 256})
# CTB says black -> resolve_color() returns #ffffff
```

On a real title block this silently dropped 4 ACAD_TABLE/INSERT entities from the plot.
Plotting the same drawing and CTB through AutoCAD 2025 renders them black.

`resolve_lineweight()` already handles ACI 7 correctly, which is what suggested to me
this is an oversight in the color path rather than intended behaviour.

## Cause

Two places bypass the plot style table for ACI 7:

- `RenderContext._aci_to_true_color` — direct entity color
- `RenderContext.resolve_layer_properties` — sets `has_aci_color_7`, which later makes
  `LayerProperties.get_entity_color_from_layer` return the layout foreground instead of
  the layer color (the BYLAYER path)

I understand the special case comes from #227: ACI 7 has to follow the background
(white on dark, black on light). That is right **while ACI 7 is ambiguous** — but a
plot style table is an explicit instruction that resolves the ambiguity, so it should
win.

## Change

Both places now consult the table first and fall back to the background-dependent
default only when the table expresses no opinion for ACI 7.

Detecting "expresses no opinion" needed a small addition: `_load_plot_style_table()`
fills every entry that `has_object_color()` with the default AutoCAD palette, so after
loading the table can no longer tell the two cases apart. The set of ACI values the
table really overrides is now recorded during that same pass, as `forced_colors`.

That method can run more than once over the same table (`set_current_layout()`,
`from_viewport()`), so the detection is guarded to happen on the first pass only —
without that guard every ACI ends up looking "forced" on the second run. This is
covered by `test_plot_style_without_color_keeps_background_behavior`.

## Tests

Four new tests in `TestPlotStyleOverridesACIColor7`:

- explicit ACI 7 uses the plot style color
- BYLAYER on layer `0` uses the plot style color
- other ACI values are unaffected
- a table with no color for ACI 7 keeps the current background-dependent behaviour
  (both dark and light background)

Verified that the first two **fail without this change** and pass with it.

Full suite: **7508 passed, 71 skipped, 1 xfailed** (7504 before, plus the 4 new ones).
No existing test changed behaviour — in particular `TestResolveLayerACIColor7`, which
covers the no-CTB case from #227, still passes untouched.

On the real drawing that surfaced this, the title block now resolves 159/159 entities
to black instead of 155 black + 4 white.


mozman:
FYI: This was never planned and never will be a perfect DXF viewer! 😉

PR #1403, opened 2026-08-10
Raise DXFStructureError on an overflowing integer tag value

eeshsaxena:
### Problem

`tag_compiler()` in `lldxf/tagger.py` falls back to `int(float(x.value))` for integer group codes whose value is stored as a float (ProE writes ints as floats). It catches `ValueError`, but a value like `1e999` parses to `float('inf')`, and `int(inf)` raises `OverflowError`, which escapes as a raw exception when reading a malformed DXF file:

```python
import io
from ezdxf.lldxf.tagger import tag_compiler, ascii_tags_loader
list(tag_compiler(ascii_tags_loader(io.StringIO('62\n1e999\n0\nEOF\n'))))
# OverflowError: cannot convert float infinity to integer
```

### Fix

Catch `OverflowError` alongside `ValueError` and raise `DXFStructureError`, consistent with how other invalid tag values are reported.

### Test

Added `test_ext_overflow_int_value` in `test_042_tags_loader.py`. It fails with `OverflowError` before this change and passes after; the full `test_042_tags_loader.py` suite passes (23 tests).


mozman:
`Ezdxf` has a special module `ezdxf.recover` to read crappy DXF files.

https://ezdxf.mozman.at/docs/drawing/recover.html#module-ezdxf.recover

Please test if this module loads your problem DXFs. If not create a new PR.

mozman:
Addition: If you add a solution, please add it to the `ezdxf.recover` module. The regular loaders should not be littered with special cases to stay as fast as possible for well formed DXF files.

PR #1404, opened 2026-08-13
Expose base math types to static type checkers

eevleevs:
Expose Vec2, Vec3, and Matrix44 from ezdxf.math to static type checkers while preserving the runtime C-extension implementations via TYPE_CHECKING. The pure-Python _ctypes module inspected by type checkers does not export these names, causing imports such as from ezdxf.math import Vec3 to resolve as Unknown. Validation: python -m py_compile src/ezdxf/math/__init__.py.


mozman:
These types are imported in `ezdxf.math.__init__.py` from `ezdxf.math._ctypes.py`

PR #1405, opened 2026-08-13
Prefer typed math implementations during type checking

eevleevs:
## Problem
`ezdxf.math._ctypes` already exports `Vec2`, `Vec3`, and `Matrix44`, but its runtime implementation selection causes static type checkers to see `Unknown | Vec3`: the Cython branch has no Python typing information, while the pure-Python fallback is typed.

## Fix

Guard the Cython branch with `not TYPE_CHECKING`. Type checkers therefore select the typed pure-Python fallback, while runtime behavior remains unchanged: Cython implementations are still used whenever `USE_C_EXT` is true.

## Validation

- `python -m py_compile src/ezdxf/math/_ctypes.py`
- `git diff --check`


mozman:
FYI: the current state of the repository is working:
```bash
(.venv) C:\Source\ezdxf.git
>pip list | grep mypy
mypy                          2.3.0
mypy_extensions               1.1.0

(.venv) C:\Source\ezdxf.git
>make mypy
mypy --ignore-missing-imports -p ezdxf
Success: no issues found in 343 source files
```

eevleevs:
Added commit 95ee6796b with the follow-up type fixes. This PR changes mypy's view of ezdxf.math._ctypes so it uses the typed Python implementations instead of treating the C extensions as unresolved/Any. That exposed type errors that were already present in the codebase; the follow-up fixes address those preexisting issues. Runtime C-extension selection is unchanged.\n\nValidation:\n- python -m mypy --ignore-missing-imports src/ezdxf\n- No issues found in 341 source files\n- 7121 tests passed, 30 skipped, 1 xfailed\n\nDrawing tests were excluded because PIL is not installed locally.

mozman:
Just NO!

eevleevs:
Quite rude, care to explain why? Consider that this shows you have many suppressed inconsistencies in your codebase.

PR #1406, opened 2026-08-17
Raise DXFStructureError instead of OverflowError on out-of-range int tag

eeshsaxena:
The tag compiler already tolerates float-formatted integer values (the ProE case) by falling back to `int(float(x.value))`, but that fallback only catches `ValueError`. A value like `1e999` parses as `inf`, so `int(float(...))` raises `OverflowError` and the whole file load aborts with a raw exception instead of a `DXFStructureError`. Catching `OverflowError` alongside `ValueError` there lets it degrade to a structure error like every other malformed int tag.


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`Raise DXFStructureError instead of OverflowError on out-of-range int tag`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining ezdxf!

mozman:
`Ezdxf` has a special module `ezdxf.recover` to read crappy DXF files.

https://ezdxf.mozman.at/docs/drawing/recover.html#module-ezdxf.recover

Please test if this module loads your problem DXFs. If not create a new PR.

Issue #1399, opened 2026-07-19
HATCH.transform() re-rotates the pattern lines by pattern_angle on every transform (a pure translation rotates the pattern)


tuxiasumari:
### Describe the bug

`DXFPolygon.transform()` rotates the realized pattern lines by the **absolute** `dxf.pattern_angle` on every call, instead of by the rotation delta of the transformation. A pure translation therefore visibly re-orients the hatch pattern, and the error is cumulative: translate twice, and the pattern is rotated by `2 × pattern_angle`.

In `transform()` (master, `src/ezdxf/entities/polygon.py`):

```python
angle = ocs.transform_deg_angle(self.dxf.pattern_angle)
relative_factor = ocs.transform_length((1, 0, 0))
self.pattern.scale(relative_factor, angle)
```

`Pattern.scale(factor, angle)` applies its `angle` argument **relative** to the already-realized lines (it rotates the current definition), but `angle` here is the **absolute** new pattern angle. For a pure translation `transform_deg_angle(pattern_angle) == pattern_angle`, so the lines get rotated by the full pattern angle again. Compare with `set_pattern_angle()`, which correctly passes the delta:

```python
self.pattern.scale(angle=angle - dxf.pattern_angle)
```

Note the *scale* half of this was already made relative (`relative_factor`, I believe with #1392) — the angle needs the same treatment:

```python
old_angle = self.dxf.pattern_angle
angle = ocs.transform_deg_angle(old_angle)
self.pattern.scale(relative_factor, angle - old_angle)
```

### Minimal repro

```python
import ezdxf
from ezdxf.math import Matrix44

doc = ezdxf.new("R2018")
msp = doc.modelspace()
h = msp.add_hatch()
h.paths.add_polyline_path([(0, 0), (10, 0), (10, 10), (0, 10)])
h.set_pattern_fill("ANSI31", scale=1.0, angle=30.0)

print("line[0].angle before:          ", h.pattern.lines[0].angle)
h.transform(Matrix44.translate(100, 0, 0))
print("after translate(100,0):        ", h.pattern.lines[0].angle)
h.transform(Matrix44.translate(100, 0, 0))
print("after another translate(100,0):", h.pattern.lines[0].angle)
```

Output with current master (1.4.5b0) and 1.4.4:

```
line[0].angle before:           75.0
after translate(100,0):         105.0   (expected 75.0)
after another translate(100,0): 135.0   (expected 75.0)
```

[...]

Issue #1392, opened 2026-06-15
Bug: DXFPolygon.transform() fails to apply rotation and translation to hatch pattern base_point inside nested INSERT


origami7:
### Description

**Describe the bug**
When a `HATCH` entity is nested inside an `INSERT` (Block Reference) that has a rotation angle (e.g., 180 degrees), the `Frontend` rendering pipeline fails to render any hatch lines (outputting an empty/transparent filling). 

Through deep debugging, we found that the root cause lies in `DXFPolygon.transform()` (located in `src/ezdxf/entities/polygon.py`). While it correctly transforms the boundary paths, it **fails to properly apply the rotation and translation matrices to the hatch pattern's `base_point`**, applying only the pattern scale factor instead. This causes the pattern origin to fly millions of units away from the actual boundary, causing the internal ray-casting algorithm (`hatch_lines()`) to yield 0 line segments due to massive signed distance divergence.

### Mathematical Proof (Debug Evidence)

Here is the exact state comparison captured from a real failing case where the `INSERT` has a **180° rotation** and a translation matrix $T = (962.99, 969.88)$:

#### 1. Boundary Paths (Correctly Transformed)
* **Original vertex in block definition**: `(-64.92, 173.74)`
* **Transformed vertex in WCS**: `(1027.91, 796.14)` 
* *Calculation*: $180^\circ \text{ rotation} + T \implies (962.99 - (-64.92), 969.88 - 173.74) = (1027.91, 796.14)$  **(✓ Correct)**

#### 2. Pattern Base Point (Buggy Transformation)
* **Original pattern base_point**: `(-38543.80, -6670.83)`
* **Expected WCS base_point (after 180° rotation + Translation + Scale 200)**: 
  $$\begin{aligned} X &= (-(-38543.80) + 962.99) \times 200 = 7,901,358 \\ Y &= (-(-6670.83) + 969.88) \times 200 = 1,528,142 \end{aligned}$$
* **Actual Buggy Result produced by `transform()`**: `(7708760.44, 1334166.08)`
* *Dissection*: The code mistakenly skipped the rotation matrix and translation, evaluating to roughly $\text{base\_point} \times \text{scale}$, shifting the emission origin away by a delta of millions of units.

#### 3. Consequence on `hatch_lines()`
* **Signed distance from boundary to baseline**: `-4,507,355`
* **Pattern line spacing (`normal_distance`)**: `5,000`
[...]

Issue #1391, opened 2026-06-10
Hatch pattern scale issue


thbeu:
I wonder what is the correct hatch pattern scale for the STAR pattern of 1CF of sample file https://github.com/jscad/sample-files/blob/master/dxf/dxf-parser/entities.dxf. I can see differences in scale between ezdxf and dxfViewer (WLX for Total Commander),

<img width="644" height="430" alt="Image" src="https://github.com/user-attachments/assets/6eb62607-381e-48e6-a61f-fadaa65e1eb0" />

PR #1407, opened 2026-08-19
Fix #1402: translate hatch pattern base points in DXFPolygon.transform()

### LINKED ISSUE #1402: DXFPolygon.transform() never translates hatch pattern base points — pattern phase shifts relative to the boundary under any translation (still on master, cf. #1392)
## Description

`DXFPolygon.transform()` applies the full affine transformation to the boundary paths, but the hatch pattern only receives the **linear part** (scale + rotation) — the translation is never applied to the pattern line base points:

```python
# src/ezdxf/entities/polygon.py, DXFPolygon.transform (master, 1.4.5b0)
self.paths.transform(ocs, elevation=elevation)   # boundary: full affine
...
if self.pattern:
    relative_factor = ocs.transform_length((1, 0, 0))
    rotation_angle = ocs.transform_deg_angle(0)
    self.pattern.scale(relative_factor, rotation_angle)   # linear part only
```

`Pattern.scale()` → `tools/pattern.scale_pattern()` only rotates/scales `base_point` about the origin; there is no translation term anywhere on the pattern side.

Since hatch lines are generated at integer multiples of `offset` from `base_point` (per #769: "resetting the pattern origin works by resetting the base point of all pattern definition lines"), any translated hatch ends up with its pattern **phase** shifted by `(translation · line_normal) mod offset` relative to its boundary. The geometry (line count, pitch, direction) stays correct — only the phase is wrong, which makes the defect easy to miss visually but systematic.

## Minimal repro (state-level, render-independent — master 1.4.5b0)

```python
import ezdxf
from ezdxf.math import Matrix44

doc = ezdxf.new()
msp = doc.modelspace()
h = msp.add_hatch()
h.set_pattern_fill("ANSI31", scale=10)
h.paths.add_polyline_path([(0, 0), (100, 0), (100, 100), (0, 100)], is_closed=True)

h.transform(Matrix44.translate(37.5, 11.0, 0))

print(h.paths.paths[0].vertices[0])      # (37.5, 11.0, 0.0)  — boundary moved
print(h.pattern.lines[0].base_point)     # (0.0, 0.0)         — pattern did not
```

Output on current master:

```
boundary vertex: (0, 0, 0.0) -> (37.5, 11.0, 0.0)  (moved 37.5, 11.0)
pattern base_point: (0.0, 0.0) -> (0.0, 0.0)        (moved 0.0, 0.0)
```

Expected: `base_point` is transformed with the same OCS mapping as the boundary vertices (as a *point*, not a direction), so the pattern keeps its position relative to the boundary — which is what AutoCAD does when a block containing a hatch is inserted or an entity is transformed.

## Relation to #1392

#1392 reported the same root cause (rotation + translation missing on `base_point`, observed via empty fills under a 180° INSERT) and was closed after a sample rendered OK, but the translation gap is still present on master — the repro above shows it at the entity-state level, independent of any rendering path.

## Real-world impact (measured)

On an AutoCAD-authored drawing where hatches live inside block definitions and are placed by translated INSERTs (flattened via `ezdxf.xref` / virtual entities → `transform()`), the rendered pattern phase deviated from AutoCAD's own plot by **+0.326 of the pattern pitch** (117 drawing units at pitch ≈ 359). Patching `transform()` to also map `base_point` through the boundary's OCS transform reduced the deviation to **+0.004 pitch**, matching the AutoCAD plot; in a 1,254-window plot-fidelity A/B against AutoCAD output the patch changed 15 verdicts, all improvements, no regressions.

## Suggested fix

In `DXFPolygon.transform()`, capture the pre-transform `base_point` of each pattern line and, after the existing `pattern.scale(...)` call, overwrite `base_point` with the point-mapped original (e.g. `ocs.transform_vertex(Vec3(bp.x, bp.y, elevation))` reduced to 2D). Mapping the original point once avoids double-applying the linear part that `pattern.scale()` already handled.

Environment: reproduced on master (1.4.5b0) and 1.4.4, CPython 3.11, Linux x64.

youdie006:
## Repro (state-level)

Under any translating transform, `DXFPolygon.transform()` moves the HATCH/MPOLYGON boundary but leaves the pattern line `base_point` where it was, so the pattern phase drifts relative to the boundary:

```python
import ezdxf
from ezdxf.math import Matrix44, Vec2

hatch = ezdxf.new().modelspace().add_hatch()
hatch.paths.add_polyline_path([(0, 0), (10, 0), (10, 10), (0, 10)])
hatch.set_pattern_fill("ANSI31", scale=1.0)
hatch.pattern.lines[0].base_point = Vec2(0, 0)   # anchored on the boundary origin

hatch.transform(Matrix44.translate(37.5, 11.0, 0))

# boundary vertex (0,0) correctly moves to (37.5, 11.0)
# BUT hatch.pattern.lines[0].base_point stays at (0, 0)   <-- bug
```

## Cause

`transform()` applies the full affine transform to the boundary paths, but hands the pattern only the linear part (scale + rotation) via `self.pattern.scale(relative_factor, rotation_angle)`. `PatternLine.base_point` is a position, not a direction, so it also needs the translation -- which is never applied.

This completes the translation half of the sibling issue #1392: the rotation half was fixed in #1399 and the scaling half in #1391, but pattern base points were still never translated.

## Fix

Capture each line's original `base_point`, run `pattern.scale(...)` unchanged, then overwrite each base point by mapping the ORIGINAL point once through the same `OCSTransform` the boundary vertices use (`ocs.transform_vertex(Vec3(bp.x, bp.y, elevation))`). Mapping the original point with the full transform avoids double-applying the linear part that `scale()` already handled. Scaling and rotation behaviour is unchanged; only the missing translation of `base_point` is added.

## Tests

Added three state-level regression tests in `tests/test_02_dxf_graphics/test_229c_hatch_transformations.py`:

- pattern base point moves by the same translation as a boundary vertex (the repro above),
- every non-zero base point across all pattern lines moves by the same offset,
- a base point coincident with a boundary vertex lands on the exact same transformed point under a combined rotation + translation (guards against double-applying the linear part).

Red/green verified: the three tests fail on current `master` and pass with the fix; the full `-k "pattern or hatch"` suite is green with no regressions (195 passed, 3 skipped, C-extension built).

Fixes #1402. Thanks to @dwscat20 for the isolated, state-level reproduction.

---
This contribution was prepared with AI assistance and reviewed by me before submission.

### DIFF
--- a/src/ezdxf/entities/polygon.py
+++ b/src/ezdxf/entities/polygon.py
@@ -408,7 +408,24 @@ class DXFPolygon(DXFGraphic):
 
             # rotation relative to current state:
             rotation_angle = ocs.transform_deg_angle(0)
+
+            # Capture the original base points before scaling/rotating the
+            # pattern. Pattern.scale() applies only the linear part (scale and
+            # rotation) of the transformation to the base points, it never
+            # translates them, so the pattern phase drifts relative to the
+            # boundary under any translation (issue #1402).
+            original_base_points = [line.base_point for line in self.pattern.lines]
             self.pattern.scale(relative_factor, rotation_angle)
+
+            # Translate the base points by mapping each original base point once
+            # with the full transformation, exactly like a boundary path vertex
+            # (see PolylinePath.transform). Mapping the original point with the
+            # full affine transform avoids double-applying the linear part that
+            # Pattern.scale() already handled.
+            for line, base_point in zip(self.pattern.lines, original_base_points):
+                line.base_point = ocs.transform_vertex(
+                    Vec3(base_point.x, base_point.y, elevation)
+                ).vec2
 
             # The pattern_scale factor has to be applied to the base pattern to get the
             # final scaling. This is important for CAD applications, not for the rendering
--- a/tests/test_02_dxf_graphics/test_229c_hatch_transformations.py
+++ b/tests/test_02_dxf_graphics/test_229c_hatch_transformations.py
@@ -239,3 +239,62 @@ def test_translation_does_not_change_pattern_line_angle():
 
     hatch.transform(Matrix44.translate(100, 0, 0))
     assert hatch.pattern.lines[0].angle == pytest.approx(pattern_line_angle)
+
+
+def test_translation_moves_pattern_base_point_like_a_boundary_vertex():
+    # Regression test for issue #1402: DXFPolygon.transform() translated the
+    # boundary paths but left the hatch pattern line base points untouched, so
+    # the pattern phase drifted relative to the boundary under any translation.
+    hatch = Hatch()
+    path = hatch.paths.add_polyline_path([(0, 0), (10, 0), (10, 10), (0, 10)])
+    hatch.set_pattern_fill("ANSI31", scale=1.0)
+    # anchor the first pattern line base point at the boundary origin (0, 0)
+    hatch.pattern.lines[0].base_point = Vec2(0, 0)
+
+    offset = Vec2(37.5, 11.0)
+    hatch.transform(Matrix44.translate(offset.x, offset.y, 0))
+
+    # the boundary vertex (0, 0) maps to the translation offset ...
+    vx, vy, _ = path.vertices[0]
+    assert Vec2(vx, vy).isclose(offset)
+    # ... and the pattern base point must move by the same translation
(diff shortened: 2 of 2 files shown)


mozman:
This was never planned and never will be a perfect DXF viewer! 😉

PR #1409, opened 2026-08-27
Refactor dimension rendering logic

origami7:
fix: restore correct dimension geometry when exploding block references
When exploding a block INSERT containing DIMENSION entities, ezdxf re-renders
each DIMENSION's anonymous geometry block (*D###) via LinearDimension.
Three bugs caused corrupted geometry after explode:
1. oblique_angle default was hard-coded to 90 instead of 0
2. parallel ray intersection raised ParallelRaysError instead of fallback
3. text_height was incorrectly scaled by dimscale (304.8) producing
   100,000+ unit char_height instead of the correct 330 units


origami7:
[test_11_ylx.zip](https://github.com/user-attachments/files/31494534/test_11_ylx.zip)

PR #1413, opened 2026-09-05
Add ascii_tag_compiler(), a fused ASCII tag loader [+15% dxf reading speed]

toporivskiy:
Loading a DXF document ran the input through two chained generators: ascii_tags_loader() built a DXFTag for every group code/value pair and tag_compiler() threw that tag away to build a second one with the value cast to its Python type. For an 83MB DXF with 1.42M tag pairs that is 2.7M generator hand-offs and 2.7M objects to produce 1.29M tags.

ascii_tag_compiler() does both steps in a single pass over the same readline() stream. Measured on that file: loading the tags takes 0.56s -> 0.40s (-28%), **loading the whole document 1.05s -> 0.91s (-14%)**.

The two-stage pipeline is untouched and stays the way to compile tags from other sources (recover, binary DXF, the DXF browser add-on). Drawing.read() and iterdxf.modelspace() use the fused loader, both only ever fed it a text stream.

The new loader is behavior-compatible down to the exception messages, which is why it keeps two line counters: one counts the lines read from the stream like ascii_tags_loader() does, the other counts the lines of the tags handed to the compiling stage like tag_compiler() does - the two differ as soon as comment tags are filtered out.

test_044_ascii_tag_compiler.py compares both implementations tag by tag over 19 hand-written samples (with and without comment filtering) and all 45 files in examples_dxf/, including the raised exceptions and the line numbers in their messages.


mozman:
I don't you how you profiled this PR, but in my experience the bottleneck of the loader is loading from disk. Therefore, I assume the initial load is slow regardless of you PR, and subsequent loads often do not happen in real world applications.

